PR #11, opened 2017-08-21
Add placeholder for the reports page

msanroman:
### Purpose

To add a placeholder for the reports page.


packages/server/index.js
```diff
@@ -64,6 +64,8 @@ app.post('/rpc', (req, res, next) => {
 
 app.get('/health-check', controller.healthCheck);
 
+app.get('/*', (req, res) => res.send(html));
```
federicoweber: great catch on this :100:

PR #15, opened 2017-08-25
Add ProfileSelector

federicoweber:
### Purpose
To manage and select profiles for a given social network.

### Notes
- we are missing Instagram color, here is a [PR for it in Buffer Components](https://github.com/bufferapp/buffer-publish/pull/138).
- it replaces 'SELECT_PROFILE` actions and dispatcher in `nav-sidebar`
- this is temporarily rendering `ProfileSelector` at the very top of the Insights page, should really be in the profileHeader, will move it there once we have it
- It adds in a basic version of the ProfileSelectorDropdown, will add all those in a separate PR so that we can use the reducers and actions in the meanwhile
    - has minimal styling
    - missing the filter functionality and manage button
    - need to explore options to handle super long lists of profiles, while maintaining usability.
#### Staging Deployment


federicoweber:
Hey @msanroman, before merging this in, I would love to catch up with you around this race condition. https://github.com/bufferapp/buffer-analyze/blob/f62c4d0e406b3931a03b5d7906a6a40c2fe74081/packages/nav-sidebar/middleware.js#L25-L29 

Added a `setTimeout` to move quickly, but it's far from the ideal solution. 
The race condition is happening cause `NavSidebar` it selecting a profile immediately after fetch, but `ProfileSelector` is yet to update its own profile collection with the fetched data.   

I think we can simplify the flow if we pick just one location for the `profiles` _collection_.  Both `NavSidebar` and  `ProfileSelector` could be a good home for it, but as we also need to select the first profile anyway, I have a hunch that a good solution could be to move it all to `ProfileSelector` while maintaining the fetch in  `navSidebar` after the login.
Looking forward to brainstorming on that with you :smile:.

PR #19, opened 2017-08-30
Profile Header Package

tiggreen:
### Purpose
This PR implements Profile Header package.

image 2017-08-30 at 1 24 06 pm

#### Staging Deployment


.yarnclean
```diff
@@ -0,0 +1,42 @@
+# test directories
```
tiggreen: Hmm, do I need this file at all? Maybe I've run `yarn clean` instead of `yarn run clean`?

packages/profile-header/components/ProfileHeader/index.jsx
```diff
@@ -0,0 +1,143 @@
...
+    case 'twitter':
+      icon = <CircleTwitterIcon color={service} size={{ width: '100%', height: '100%' }} />;
+      break;
+    case 'facebook':
+      icon = <CircleFacebookIcon color={service} size={{ width: '100%', height: '100%' }} />;
+      break;
+    case 'instagram':
+      // TODO using this as it's the only red we have
+      // but we need to change it to match instagram color
+      icon = <CircleInstagramIcon color={'torchRed'} size={{ width: '100%', height: '100%' }} />;
```
tiggreen: If we already have the Instagram color I can use that. I've taken `SocialIcon` from Federico's work.

packages/profile-header/components/ProfileHeader/index.jsx
```diff
@@ -0,0 +1,143 @@
...
+  );
+};
+
+ProfileHeader.propTypes = {
+  profile: PropTypes.shape({
+    service: PropTypes.string.isRequired,
+    username: PropTypes.string.isRequired,
+    avatarUrl: PropTypes.string.isRequired,
+    id: PropTypes.string.isRequired,
+  }),
```
tiggreen: Is this is a big deal that I haven't defined `profile` prop as required?

packages/profile-header/styles.less
```diff
@@ -0,0 +1,83 @@
+@base: 10px;
+
+.sectionProfileAvatar {
```
tiggreen: Most of the styles I've taken from our existing Overview Tab components. I'm not sure if I supposed to :) Would love your advice on styling.

.yarnclean
msanroman: Ah, that might be the case, yup! I'd say we're good removing the file 😄

packages/profile-header/components/ProfileHeader/index.jsx
msanroman: It isn't, but I'd add it in `ProfileHeader.defaultProps` (maybe set it to `null`?)

packages/profile-header/components/ProfileHeader/index.jsx
msanroman: I'd move `SocialIcon` to the `shared-components` package (it's available in master, you might want to merge) and reference that one from this package and the `profile-selector`, in order to prevent duplication.

Let me know if I can help with the move!

packages/profile-header/components/ProfileHeader/index.jsx
```diff
@@ -0,0 +1,143 @@
...
+      {icon}
+    </div>
+  );
+};
+
+SocialIcon.propTypes = {
+  service: PropTypes.oneOf(['twitter', 'facebook', 'instagram']).isRequired,
+};
+
+export const ProfileBadge = ({ avatarUrl, service }) => {
```
msanroman: Same thing as I mentioned on `SocialIcon`, I feel these are shared components between several of our packages, so I'd move it to `shared-components` 😄

packages/profile-header/components/ProfileHeader/index.jsx
```diff
@@ -0,0 +1,143 @@
...
+      <div className={styles.profileAvatarWrapper}>
+        <ProfileBadge avatarUrl={profile.avatarUrl} service={profile.service} />
+        <div className={styles.sectionProfileAvatarDetails}>
+          <span className={styles.profileAvatarDetailHeader}>
+            <a href={profile.service_link} rel="noopener noreferrer" target="_blank">
+              <Text size="small">{profile.username}</Text>
+            </a>
+          </span>
+          <span className={styles.profileAvatarDetailSubHeader}>
+            <Text size="extra-small" weight="bold">{formatFollowerCount(followersCount)} {followersWording}</Text>
```
msanroman: Instead of using `formatFollowerCount` maybe we can wrap this number in a `TruncatedNumber` component? It does more formatting that just `0,0`, so not sure if that'd be desired. We could always create a `ThousandSeparatedNumber` that just does that formatting, maybe?

packages/profile-header/components/ProfileHeader/index.jsx
tiggreen: Oh la la, of course 🗡 That makes sense! I'll move those 2 components to `shared-components` 😄

packages/profile-header/styles.less
msanroman: This looks fine for me. I feel we'll decide on styling by the end of the cycle (let's chat about that on the retreat!).

tiggreen: 💯 YAAAAS! I also think for now we maybe should not think very strict about styling! We can always adjust it later!

tiggreen:
Hey @msanroman and @federicoweber I've moved `ProfileBadge` and `SocialIcon` into `shared-components` and referenced them in `profile-selector` and `profile-header` packages!

PR #20, opened 2017-08-30
Date picker

msanroman:
## Purpose

To add a date picker package that handles all the data flow and state for the selected date range, as well as exposing the UI component with presets and a calendar for selecting a custom date range.

### Notes

+ Any component that needs to be informed about the date selected can consume that state with `state.date.startDate` and `state.date.endDate`. If the date changes, a `SET_DATE_RANGE` actionType (from `@bufferapp/analyze-date-picker`) will be triggered with the new `startDate` and `endDate` pair.
+ A new package named `shared-components` has been created to hold components that might be used between many packages. The first of them is `PeriodPhrase`, which receives a start and end date and formats the phrase for the period used in analytics headings such as "Performance for `the selected 7 days`/`yesterday`/`today` compared to the previous period".


packages/summary-table/middleware.js
```diff
@@ -1,13 +1,39 @@
 import { actions } from '@bufferapp/async-data-fetch';
 import { actionTypes as profileActionTypes } from '@bufferapp/analyze-profile-selector';
+import { actionTypes as dateActionTypes } from '@bufferapp/analyze-date-picker';
 
-export default ({ dispatch }) => next => (action) => { // eslint-disable-line no-unused-vars
+const getProfileIdFromRoute = (state) => {
```
federicoweber: Hey @msanroman, how do you feel about storing this in the `NavSidebar` state? In this way, we can avoid redoing the URL parsing in here. 
Other option could be to pass the `profileId` as a prop to the date picker so that it can be provided to the `SET_DATE_RANGE` action.

packages/date-picker/components/DatePicker/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+        endDate={endDate}
+        handleClick={() => (isOpen ? close() : open())}
+      />
+      { !loading &&
+        <DatePickerDropdown isOpen={isOpen}>
+          <DatePickerForm {...props} />
+        </DatePickerDropdown>
+      }
+
+      <div tabIndex="0" role="button" onClick={close} className={clickCatcherClass} />
```
federicoweber: Super small one, I noticed that it's possible to open the profile selector drop-down when the date picker is open, and it will result in a bit of overlap. I think setting this to `z-index: 1;` should fix it :smile: .

PR #30, opened 2017-09-06
Automatic async data fetch performance tracking

federicoweber:
### Purpose
to ensure we are tracking performances for all RPC endpoints, we are now tracking all async data fetch request automatically.

### Notes
To keep the middleware cases declarations Redux friendly, this is using `RegExp.test` method instead of passing the type variable as a `switch` argument.

#### Staging Deployment


federicoweber:
Thanks @tiggreen, will work on a dashboard soon :smile:.

PR #32, opened 2017-09-07
Add charts to Average Table

federicoweber:
### Purpose
to add daily average chart to the Average Table

20170908_163041-screenshot

#### Staging Deployment


federicoweber:
Hey @tiggreen this is still a WIP, so no rush to review, just wanted to put this on your radar :smile:.

federicoweber:
hey @msanroman  and @tiggreen, 
Snapshot testing is not playing that well with Highcharts, so to make it work I'm using the `shallowSnapshot` option. That works well as it prevents the snapshot tests to process sub components but the drawback is that it's decreasing tests coverage. 

We can probably solve that by adding in a few stories to cover our own Components, but  I'm also looking to see if there is a better option to make this work without the need to use Shallow.

federicoweber:
End up solving Highcharts testability with mock instead of shallow rendering :smile:.

federicoweber:
Great catch thank you @tiggreen.
Fixed that, going to :shipit:.

PR #33, opened 2017-09-11
Top Posts Table (without the header)

tiggreen:
### Purpose

This PR implements the Top Posts Package for Buffer Analyze. 
**For now I'm displaying only the top 10 posts. Next week when I'm back I'll add the drop-down sorting functionality as well as the order bar in the header.**

See the screenshot below: 

image 2017-09-13 at 2 16 38 pm

#### Staging Deployment


packages/top-posts-table/styles.less
```diff
@@ -0,0 +1,329 @@
+
+@base: 10px;
```
tiggreen: Again, I've copied the styles from buffer-web.

packages/web/components/OverviewTab/index.jsx
```diff
@@ -5,12 +5,14 @@ import { connect } from 'react-redux';
...
 import Divider from '@bufferapp/components/Divider';
 import AverageTable from '@bufferapp/average-table';
+import TopPostsTable from '@bufferapp/top-posts-table';
 
 const OverviewTab = ({ match }) => (
   <div>
     <SummaryTable profileService={match.params.service} />
     <Divider marginTop="1rem" marginBottom="1rem" />
     <AverageTable />
+    <TopPostsTable selectedProfileId={match.params.id} profileService={match.params.service} />
```
tiggreen: We need `selectedProfileId` to get the `timezone` of the selected profile.

packages/top-posts-table/components/TopPostsTable/index.jsx
```diff
@@ -0,0 +1,160 @@
...
+    const postMetric = cloneDeep(postMetricConfig);
+
+    postMetric.value = postStatistics[postMetric.apiKey];
+    post.statistics[postMetricConfig.key] = postStatistics[postMetric.apiKey];
+  });
+
+  return post;
+}
+
+function getMaxMetricValue(posts, metrics) {
```
msanroman: I'm wondering if this is something we could do on the RPC endpoint already, and prevent data transformations on the views 🤔

packages/top-posts-table/components/TopPostsTable/index.jsx
```diff
@@ -0,0 +1,160 @@
...
+import { metricsConfig } from '../../metrics';
+
+const gridContainer = {
+  minHeight: '12rem',
+  position: 'relative',
+  margin: '1rem 0 1.5rem',
+};
+
+
+function mergePostMetrics(post, postMetrics) {
```
msanroman: This as well. I know it was handled this way on buffer-web, but it might be a good chance for us to think if this is the best way to handle things. Labels can be imported already on the RPC endpoint, and the colors could be decided by giving a CSS class (or style) for each apiKey (separating further presentation from business logic).

Also, I think it would help us to reduce to a bare minimum config files like `metrics.js`.

packages/top-posts-table/components/TopPostsTable/index.jsx
tiggreen: hey @msanroman I think If we want to move this to RPC endpoint we will need to pass `profileService` as an argument to the endpoint so we can choose the appropriate `metricsConfig`. Is that fine?

tiggreen:
Okay, I've addressed all the styling changes that @msanroman pointed out and in my next Top Posts Header PR I'll address the other improvement comments! Going to merge it now 👍 !

PR #34, opened 2017-09-13
PNG Export

msanroman:
### Purpose

To allow users to export the current view charts and components as PNG, bundled in a ZIP file.

### Notes

The package that brings the export button also has a middleware that manages the DOM manipulation (and ZIP file creation) needed for handling the export. All the other components middlewares are able to listen to the export start action and report back themselves, asking for the export package to store that chart as one for exporting.

The `generatePNGFromChart` method checks if each ID provided belongs to a valid DOM component, so any middlewares that are triggering the action but their components are not available in the current view will be filtered out in here (this shouldn't have any performance impact).

For a new component that we want to export as a PNG, we need to add this on their middleware:

```javascript
    case exportActionTypes.EXPORT_TO_PNG_START:
      dispatch(exportActions.exportChart(<DOM_ID>, <FILENAME>));
      break;
```

And of course, in order to have those actionTypes and actions available, we'll need to add `@bufferapp/analyze-png-export` as a dependency for the package 😄


packages/png-export/middleware.js
```diff
@@ -0,0 +1,28 @@
...
+        .then((pngs) => {
+          downloadAsZip(exportToPNG.zipFilename, pngs);
+          store.dispatch(actions.endExportToPNG());
+          next(action);
+        });
+    case actionTypes.EXPORT_CHART:
+      if (timeoutID) {
+        clearTimeout(timeoutID);
+      }
+      timeoutID = setTimeout(() => store.dispatch(actions.processCharts()), 200);
```
federicoweber: hey @msanroman this works very well, but I was thinking we can maybe avoid the timeout by declaring the `png-export` midlleware as the last one in the chain - redux should execute those in order - and have it dispacth the `processCharts` action on `EXPORT_TO_PNG_START`. What do you think :smile:?

msanroman: Such a great idea! Just pushed a commit that should handle that. Thanks for the tip Federico! 👏 👏 👏

PR #37, opened 2017-09-17
Compare Chat Twitter and Facebook

federicoweber:
### Purpose
To add the compare Chart, covering this 3 use cases.

Facebook 
20170922_161543-screenshot

Twitter with previous period comparison
20170922_161733-screenshot

Twitter with Period total
20170922_161054-screenshot

### Notes
-  fixes a bug with Profile Selector Dropdown opening unprompted
- stores `profileService` on the `ProfileSelector` store; this can be used to fetch data etc
- Timezone fun :smile: 
Highchart is in UTC but we want to display the data according to the profile timezone; so we need to offset back the day for Highcharts if there is a negative utc difference.
https://github.com/bufferapp/buffer-analyze/blob/fda3bf8aa526e33999c94ae48995a5f02a8ccc7e/packages/compare-chart/components/Chart/index.jsx#L53-L57

#### Staging Deployment


tomredman:
Hey @federicoweber! This is looking pretty sweet on prod. There are few things I noticed, not sure if we want these fixed before merge (that's up to you and the eng team).

---

The totals and change % are always the same:

screen shot 2017-09-20 at 5 05 56 pm

---

- Clicking the dropdown arrow doesn't open the dropdown:
- Clicking anywhere on the dropdown reloads the charts:

screen recording 2017-09-20 at 05 07 pm

---

Changing profiles or dates results in the chart not loading:

screen recording 2017-09-20 at 05 11 pm


---

Tooltip copy should be more "dynamic", and we want to add the metric count too. This is what we do currently with the Twitter Overview charts on buffer.com, and I think it works well.

E.g.

>  There were no updates published, and you earned:
> - 4,232 impressions


> There were no updates published, compared 1 in the previous period, and you earned: 
> 
> - 14,232 impressions
> - 12,232 previous impressions

Some examples from our Twitter Overview:

screen shot 2017-09-20 at 5 21 51 pm

screen shot 2017-09-20 at 5 19 52 pm

Small detail here: For Twitter, we say "**... a total of 5 tweets that generated**" because we're only counting metrics from the Tweets actually sent on those days specifically. But for Facebook (and IG), we'll instead say "**... a total of 5 posts, and you earned**" because we count metrics from all posts, not just the ones sent on that day.

---

Super small one, the days shouldn't have the '.' after the number. E.g, "15. Sep" should be "15 Sep". We might try and make this better overall in the future based on the user's locale, so it's cool to leave this for now :)

screen shot 2017-09-20 at 5 25 33 pm

federicoweber:
Oh wow @tomredman thanks for the great review. Those bugs are pretty bad, sorry I missed those, definitely going to fix those before merging it in.
:100: on the tooltip.

federicoweber:
Hey @tomredman I think the last few commits addressed most of the issue.
Still need to figure out why react is rerendering the chart when opening the dropdown, will look into that.

Also noticed that the marker on the chart are not snapping on the grid, will look at that as well.

20170920_201141-screenshot

tomredman:
Thanks @federicoweber, this is much better!

> One question on the tooltip the metric countsThat is there apart from the updates count metrics. Do we want to add the count in for the count metrics as well?

No, just the first part is fine. E.g, `There was 1 post published.` And this one is _super_ tiny, but right now, when there's 1 update, the tooltip reads `There were a total of 1 post published...`. This should read `There was a total 1 post published...`

> Noticed that the marker on the chart are not snapping on the grid, will look at that as well.
>
> Also realized an interesting thing: and this is happening for all the charts, we are starting and ending the data fetching periods, one day later than for web. @msanroman, I remember we chatted about removing a date offset in the DateRange, maybe that is what is causing the discrepancy in here 😄.

Yeah, we ran into this with the original Twitter Overview as well. That time, it was related to timezone offsets... the grids are at midnight (maybe even UTC?), and the data points are relative to the user's timezone, if I recall correctly. We may need to normalize this data somehow so the data points are midnight in the timezone of the profile.

This was no doubt a bit of tricky one to wrap my head around at the time. Happy to help brainstorm if needed!

federicoweber:
Good call on UTC @tomredman that made the trick :smile:

PR #48, opened 2017-10-18
Add report flow

msanroman:
# Purpose

"Add report" flow for summary table, as well as a minimal UI for the reports list (detail not available yet).


packages/report-list/components/Report/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+`;
+
+const Date = styled.span`
+  color: #c1c1c1;
+`;
+
+const Report = ({ updated_at, name }) =>
+  <ReportListItem>
+    <Text size="large" weight="bold"><Name>{name}</Name></Text>
+    <Text><Date>{moment(updated_at, 'x').format('MMMM D, YYYY')}</Date></Text>
```
federicoweber: I think we need to use the timezone in here. 
:thinking: as the report is user-centered, maybe we can use `moment.tz.guess()` in this case, to unify all the dates for the user timezone instead of the profiles. What do you think?

packages/report-list/middleware.js
```diff
@@ -0,0 +1,20 @@
+import { LOCATION_CHANGE } from 'react-router-redux';
+import { actions } from '@bufferapp/async-data-fetch';
+
+export default store => next => (action) => { // eslint-disable-line no-unused-vars
+  switch (action.type) {
+    case LOCATION_CHANGE:
```
federicoweber: interesting naming choice; I would have gone for `PATH_CHANGE` :smile:

PR #52, opened 2017-10-23
Audience Comparison Chart

tiggreen:
This implements the comparison audience chart in `/comparisons` page for a single profile selector.

More details in this [flow card](https://app.getflow.com/organizations/369191/teams/339227/tasks/27067425)
### Review
Hey @msanroman and @federicoweber would really love your thorough review on this PR! Thank you 🤗 
#### Staging Deployment


packages/compare-chart/middleware.test.js
```diff
@@ -146,7 +146,7 @@ describe('middleware', () => {
     middleware(store)(next)(action);
     const csvData = {
       Posts: [1, 1, 1],
-      date: ['10/10/2017', '10/11/2017', '10/12/2017'],
+      date: ['10/09/2017', '10/10/2017', '10/11/2017'],
```
tiggreen: compare-chart tests were failing so this fixed on my end. Need to understand why they failed in the first place.

federicoweber: umm interesting, could be a timezone issue :thinking:

packages/nav-sidebar/index.js
```diff
@@ -10,10 +10,12 @@ export default connect(
   dispatch => ({
     onClick: (path, profileId, profileService) => {
       dispatch(push(path));
-      dispatch(profilesActions.selectProfile(profileId, profileService));
+      if (path === '/comparisons') {
```
federicoweber: I think we can do without the if in here, as the `profileService` will be undefined anyway right?

Also I think we can remove the `profileId`, as we are not using it anymore.

packages/profile-selector/index.jsx
```diff
@@ -28,10 +28,15 @@ const filterProfilesByService = (profiles, service) => (
 );
 
 const mapStateToProps = (state) => {
-  const filteredProfiles = filterProfilesByService(
-    state.profiles.profiles,
-    state.profiles.selectedProfileService,
-  );
+  let filteredProfiles = null;
+  if (state.profiles.selectedProfileService) {
```
federicoweber: super small one, how about pushing this logic to `filterProfilesByService ` ?

packages/profile-selector/middleware.js
```diff
@@ -43,15 +48,42 @@ const getProfileRoute = (state, id) => {
 export default ({ dispatch, getState }) => next => (action) => {
   switch (action.type) {
     case `profiles_${actionTypes.FETCH_SUCCESS}`:
+      // insights
```
federicoweber: Maybe we can extrapolate this into a `selectProfileByRoute` method. 

That will also enable use the match results for both the conditional and the assignment — not a big deal as regex are super fast, but I like to reuse things :smile: .

packages/profile-selector/middleware.js
```diff
@@ -43,15 +48,42 @@ const getProfileRoute = (state, id) => {
...
+        dispatch(profilesActions.selectProfile(firstProfile.id, firstProfile.service));
       }
       break;
-    case profileActionTypes.SELECT_PROFILE:
-      dispatch(push(getProfileRoute(getState(), action.id)));
+    case profileActionTypes.SELECT_PROFILE_SERVICE: {
+      const allProfiles = getState().profiles.profiles;
+      // if profile service is selected. E.g. Insights page
+      if (action.profileService) {
+        const filteredProfiles = filterProfilesByService(
```
federicoweber: same as before I think we should move this logic to `filterProfileByService` :smile:

packages/profile-selector/reducer.js
```diff
@@ -32,6 +33,12 @@ export default (state = initialState, action) => {
         selectedProfileId: action.id,
         selectedProfileService: action.profileService || state.selectedProfileService,
       });
+    case actionTypes.SELECT_PROFILE_SERVICE:
+      return {
+        ...state,
+        selectedProfileService: action.profileService,
+        selectedProfileId: state.selectedProfileId,
```
federicoweber: I think we don't need to explicit the `profileId` as it's already in the state, and we are not changing it.

federicoweber:
Noticed some snapshot tests are failing for the reports; look like a timezone issue to me. Not connected to this PR but something we need to look at :smile:.

msanroman:
Good one, I'll try to fix it on the report detail PR @federicoweber! 😄

PR #53, opened 2017-10-26
Use analytics/contextual dailyTotals

federicoweber:
### Purpose
To use the dedicated analytics/contextual.json to get all the data for Contextual compare.

#### Staging Deployment


packages/server/rpc/contextual/presets.js
```diff
@@ -1,85 +1,40 @@
...
-    //     label: 'Engagement',
-    //     key: 'engagement',
-    //     metrics: [
-    //       { keys: ['page_engagements', 'post_reach'], aggregationRule: '[0]/[1]' },
-    //     ],
-    //   },
-    // },
-    {
-      id: 3,
+  facebook: {
```
msanroman: Do we need to wrap them with the profile service? I know right now it's only available for Facebook, but maybe they would work in the exact same way for Twitter, too?

packages/shared-components/TruncatedNumber/index.jsx
```diff
@@ -17,6 +17,8 @@ const TruncatedNumber = ({ children, absoluteValue, shorterOption }) => {
     formattedNumber = numeral(number).format('0a');
   } else if (number >= 10000) {
     formattedNumber = numeral(number).format('0.0a');
+  } else if (number < 1 && number > 0) {
```
msanroman: Whoa, great catch 😄

packages/server/rpc/contextual/presets.js
federicoweber: Hey @msanroman great point, as of now we don't really need to wrap this based on the services. 

The reason I've opted for this structure is that I can see us changing some of the wording, or even presenting completely different presets, for different services. Also, this keeps the presets configuration structure similar to the metrics one, making it more predictable.

PR #55, opened 2017-10-30
get session with analyze key

hharnisc:
### Purpose

Analyze only requests the session data it needs for requests.

### Notes

Extra potentially sensitive session data is not passed around unless needed.

### Review

🤷

#### Staging Deployment

https://task-get-session-with-keys.analyze.buffer.com :smile:_


msanroman:
Thanks @hharnisc!

PR #66, opened 2017-11-10
New chart containers

moreofmorris:
### Purpose
This introduces a few new components that provide a reliable way of charts having a card container and header that is inline with the new design language. There are other changes across all the charts to do with spacing, padding, margin and using the new font Roboto.

### Notes
There are still many things to update across the board with charts. There is further refactoring of components that could happen to make things a little more consistent across the charts. I always remember a very wise engineer (called Mike 😉) that said I shouldn't do too much change in one go. So this just introduces the chart cards, without going the whole way with refactoring Titles and Headers and all sorts.

### Review
Hey there @msanroman, it'd be awesome to check this over when you get some time. From my tests it doesn't seem to mess with your awesome Reports work and all the charts still seem to function well.

#### Staging Deployment

https://task-new-chart-containers.analyze.dev.buffer.com :smile:_


moreofmorris:
Hey there @msanroman and @federicoweber, I've just updated this pull request with a fresh merge from master and fresh tests and such. Would love some eyes (and possible approval) is poss :) Thank you!

PR #68, opened 2017-11-13
Add charts to existing report

msanroman:
### Purpose

To allow adding charts to existing reports.

### Notes

I've separated the `Modal` component from the `add-report` package, thanks to @federicoweber for pointing that out on a previous PR!


packages/shared-components/Modal/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+  border-radius: 5px;
+  padding: 2.25rem 1.5rem;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%,-50%);
+  z-index: ${modal};
+`;
+
+const Modal = (props, children) => {
```
federicoweber: :heart_eyes:

PR #70, opened 2017-11-13
show contextual compare for yesterday

federicoweber:
### Purpose
To show the contextual compare when  yesterday is selected as the time period

#### Staging Deployment


federicoweber:
you are right, sorry about that @msanroman

PR #87, opened 2017-11-21
Likes and Engagement Comparison Charts

tiggreen:
### Purpose
This PR creates Likes and Engagement comparison charts. Most of the components they use are shared across all charts which makes it a breeze to add a new chart to the comparison page. 

**Please note I'll have a separate PR for the backend part for these two charts. Should be a small PR with minor additions**. 

#### Staging Deployment

### DIFF
--- a/packages/engagement-comparison-chart/.babelrc
+++ b/packages/engagement-comparison-chart/.babelrc
@@ -1,0 +1,10 @@
+{
+  "presets": [
+    "es2015",
+    "react",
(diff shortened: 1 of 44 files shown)


tiggreen:
These 2 charts are very similar to other charts that have been reviewed in the past. If it's okay I'm going to merge this and deploy. If you have any comments in this PR I'd be very happy to address those after :)

PR #89, opened 2017-11-22
fix regeneratorRuntime error

hharnisc:
### Purpose

Was getting a `regeneratorRuntime` when running the application due to an `async` polyfill missing

### Notes

Requires in the `babel-polyfill` first to fix the problem


hharnisc:
@michael-erasmus @tiggreen looks like the same test error is happening due to a date following the current date:

```
lerna ERR!     -              From: 11/14/17 - To: 11/20/17
lerna ERR!     +              From: 11/15/17 - To: 11/21/17
```

Something time related! Going to merge this one in since this it doesn't seem to be breaking anything else.

PR #93, opened 2017-11-28
Working logout

hharnisc:
### Purpose

Logout menu item in App Sidebar will logout user from Analyze.

### Notes

This logs a user out of all applications.

### Review

I moved `measureFromNavigationStart` into a common place for initialization events to be dispatched. Alternatively this could go into a package, but I don't think we have enough examples to draw a conclusion. 

#### Staging Deployment

https://task-logout-analyze.dev.buffer.com


hharnisc:
Fixed that DatePicker time dependency issue :smile:

PR #95, opened 2017-12-07
Directly send perf measures to DataDog

federicoweber:
### Purpose
To track performances in Data Dog independently from buffer web.
It also adds the tracking of first meaningful paint.

#### Staging Deployment

### DIFF
--- a/packages/profile-selector/middleware.js
+++ b/packages/profile-selector/middleware.js
@@ -2,6 +2,7 @@
 import { push, LOCATION_CHANGE } from 'react-router-redux';
 import { actions as reportActions } from '@bufferapp/report-list';
 import { actionTypes } from '@bufferapp/async-data-fetch';
+import { actions as performanceActions } from '@bufferapp/performance-tracking';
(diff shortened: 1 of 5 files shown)


packages/profile-selector/middleware.js
```diff
@@ -60,6 +61,7 @@ export default ({ dispatch, getState }) => next => (action) => {
         const firstProfile = action.result[0];
         dispatch(profilesActions.selectProfile(firstProfile.id, firstProfile.service));
       }
+      dispatch(performanceActions.measureFromNavigationStart({ name: 'firstMeaningfulPaint' }));
```
msanroman: Which was the thought process for defining the "profiles loaded" event as the first meaningful paint? :)

federicoweber: great question @msanroman :).

I opted for this as it's when the sidebar becomes useful; as we need the profiles to render the full navigation. Does it make sense?

Happy to brainstorm around other possible hooks for it :smile:.

PR #97, opened 2017-12-13
Improve dev experience 💻✨

msanroman:
### Purpose

To improve dev experience by moving webpack bundling outside the container and fixing HMR.

### Notes

To make sure you have the last version of the devtools installed, please run:

`rm -rf node_modules && yarn run clean && yarn`

After that, just remember to have always `yarn watch` running while coding :)

### DIFF
--- a/packages/server/webpack.config.dev.js
+++ b/packages/server/webpack.config.dev.js
@@ -1,4 +1,5 @@
 const webpack = require('webpack');
+const fs = require('fs');
 const config = require('./webpack.config');
 
 // NOTE: Bugsnag will not notify in local setup with current weback configuration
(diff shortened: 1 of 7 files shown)


packages/server/webpack.config.dev.js
```diff
@@ -7,11 +8,27 @@ config.devtool = 'cheap-module-eval-source-map';
...
   new webpack.HotModuleReplacementPlugin(),
+  new webpack.NamedModulesPlugin(),
 );
 
+config.devServer = {
+  hot: true,
+  publicPath: config.output.publicPath,
+  contentBase: false,
+  port: 8080,
+  host: 'analyze.local.buffer.com',
```
msanroman: Hey @hamstu! I've used `analyze.local.buffer.com` so that there are no conflicts if we needed to run together Publish and Analyze locally. :)

packages/web/index.jsx
```diff
@@ -30,8 +30,14 @@ const renderApp = (AppComponent) => {
 renderApp(App);
 
 if (module.hot) {
-  module.hot.accept('./components/App', () => {
+  module.hot.accept([
+    '../store/index.js',
+    '../store/reducers.js',
+    './components/App/index.jsx',
```
msanroman: It seems that just by requiring this we get HMR working well!

- `store/index.js` watches for middlewares
- `store/reducers.js` watches for reducers
- `./compoents/App/index.jsx` watches for any child component.

msanroman: cc: @hamstu

PR #110, opened 2018-01-09
Add data tracking

tomredman:
### Purpose
Adds automated data tracking through `@bufferapp/buffermetrics` Express & Redux middleware. This was implemented using the instructions here: https://github.com/bufferapp/buffer-js-buffermetrics#install

**TODO:**

- Verify these tracking requests are firing the correct URL. Locally this is `analyze.local.buffer.com/buffermetrics/track_action` but that doesn't seem right.
- Merge PR into `buffermetrics` that ignores Redux actions that aren't wrapped. We have a lot of fetching/performance actions that we don't want to track necessarily, and add dozens and dozens of requests to every page load.

#### Staging Deployment


packages/server/index.js
```diff
@@ -52,13 +52,15 @@ const html = fs.readFileSync(join(__dirname, 'index.html'), 'utf8')
...
-//app.use(bodyParser.json());
+app.use(bodyParser.json());
 app.use(buffermetricsMiddleware({
   name: 'Buffer-Analyze',
   debug: !isProduction,
-  trackVisits: true
+  trackVisits: true,
 }));
 
+app.post('/buffermetrics/track_action', req => req.buffermetrics.trackAction(req.body));
```
tomredman: Ahhhhhhhhh smart! 👍

djfarrelly:
@tomredman @msanroman - I've incorporated your fixes in https://github.com/bufferapp/buffer-js-buffermetrics/pull/18 and https://github.com/bufferapp/buffer-js-buffermetrics/pull/16 and have published `0.5.0-beta4`. This hopefully should work for analyze now. Can you try that package in this branch and see how it works?

packages/add-report/actions.js
```diff
@@ -1,7 +1,9 @@
-export const actionTypes = {
+import keyWrapper from '@bufferapp/keywrapper';
+
+export const actionTypes = keyWrapper('REPORT', {
   CREATE_REPORT: 'CREATE_REPORT',
```
djfarrelly: @tomredman, @msanroman With `keywapper` you can remove the right side of this assignment as it will be overwritten to `REPORT__ CREATE_REPORT` anyway. Seeing the right side of the argument to `0` will also remove characters from the final bundle reducing it's size.

packages/server/index.js
```diff
@@ -50,6 +52,14 @@ const html = fs.readFileSync(join(__dirname, 'index.html'), 'utf8')
...
 app.use(logMiddleware({ name: 'BufferAnalyze' }));
 app.use(cookieParser());
+app.use(bodyParser.json());
+app.use(buffermetricsMiddleware({
+  name: 'Buffer-Analyze',
+  debug: !isProduction,
+  trackVisits: true,
+}));
+
+app.post('/buffermetrics/track_action', req => req.buffermetrics.trackAction(req.body));
```
djfarrelly: @tomredman This isn't needed - the middleware itself handles all of the `buffermetrics` endpoints: https://github.com/bufferapp/buffer-js-buffermetrics/blob/49e7cd6e5778f1c58b09cd031185e4c7f918305b/middleware.js#L16-L38

djfarrelly: If this doesn't work let me know, because it should automatically handle the requests.

packages/store/index.js
```diff
@@ -48,6 +49,15 @@ const configureStore = (initialstate) => {
     window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ ?
       window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({}) : compose;
 
+  const buffermetricsMiddleware = createMiddleware({
+    application: 'ANALYZE',
+    metadata: (state, action) => ({
+      userId: state.appSidebar.user.id,
+      profileId: state.profiles.selectedProfileId,
+      action,
```
djfarrelly: Buda won't currently do anything with this action. It might actually cause errors as the `metadata` field in the buda protobuf message must be a flat map of key-value pairs and not have nested data structures. It'd be good to remove this to avoid any issues.

tomredman:
@msanroman - @michael-erasmus left some thoughts on setting the `host` for the metrics middleware:

> Hey Tom!
> 
> Thanks so much, that helped a lot! I had a poke around the Kubernetes logs and it seems like there is still an error on the backend when the event is being sent to the events collector. This is what I'm seeing in the logs:
> 
> **BufferMetrics: Error tracking metric: Connect Failed**

> Hey Tom and Mike!
>
> One thing I noticed that could be an issue is that we might need to pass the host of the event-collector to the buffermetrics middleware, like we do here in the marketing site: 
>
> https://github.com/bufferapp/buffer-marketing/blob/079674c3f54547a25bb56b91ccf90db8b9f8f377/index.js#L81
>
> I think the middleware is being initialized here: https://github.com/bufferapp/buffer-analyze/compare/task/add-data-tracking#diff-faf0f8cfa18ad1933342fa6ce2dfd984R53
> 
> Not sure if that will do the trick, but could we try that perhaps?

msanroman:
Cool! Just pushed that tweak @tomredman!  With this tweak, we are fully set up (hopefully 🤞!).

michael-erasmus:
Hi @tomredman and @msanroman!

Thanks for adding that tweak! Sadly I'm still not seeing any data :frowning: 

I'm going to investigate a little further, will let you know what I find!

msanroman:
Awesome, it looks like we have data tracked now!

Will look into merging this PR in a bit :)

djfarrelly:
@msanroman @tomredman - I've added the automatic lowercase conversion and a default host to hopefully help simplify future configuration. The new stable release is out for `0.5.0`: https://github.com/bufferapp/buffer-js-buffermetrics/tree/v0.5.0

PR #111, opened 2018-01-10
Added Fullstory support

tomredman:
### Purpose
Adds FullStory session tracking:

screen shot 2018-01-10 at 12 19 52 pm

#### Staging Deployment

### DIFF
--- a/packages/store/initMiddleware.js
+++ b/packages/store/initMiddleware.js
@@ -1,12 +1,25 @@
 import { actions as performanceActions } from '@bufferapp/performance-tracking';
+import {
+  actions as dataFetchActions,
+  actionTypes as dataFetchActionTypes,
+} from '@bufferapp/async-data-fetch';
 
 export default ({ dispatch }) => next => (action) => {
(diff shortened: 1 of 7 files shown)


packages/store/initMiddleware.js
```diff
@@ -1,12 +1,25 @@
...
 export default ({ dispatch }) => next => (action) => {
   next(action);
   switch (action.type) {
     case 'APP_INIT':
-      dispatch(performanceActions.measureFromNavigationStart({ name: 'init' }));
-      break;
+        dispatch(performanceActions.measureFromNavigationStart({ name: 'init' }));
+        dispatch(dataFetchActions.fetch({ name: 'user' }));
+        break;
+    case `user_${dataFetchActionTypes.FETCH_SUCCESS}`:
```
msanroman: I believe we don't really need all these tweaks, they should be included already in the app-sidebar middleware, we just need to upgrade the package version.

I can take a look at it and push those changes to this PR :)

PR #113, opened 2018-01-10
Replace "logged in" message with simple dashboard

tomredman:
### Purpose
Replace "You are logged in!" message with something simple but useful:

screen shot 2018-01-10 at 5 49 41 pm

#### Staging Deployment


tomredman:
Need to update the snapshots to pass tests 👍

PR #119, opened 2018-01-17
Improving empty states

moreofmorris:
### Purpose
This is to improve the empty states of reports, comparisons and graphs. They are not quite there in the design sense, but I wanted to get something in there before the week was out. I aim to improve these and make them more user-friendly as time goes on. 👍 

### Notes
This touches quite a few parts of the app. The reports list, add to report modal, comparisons page and the charts no data component. I didn't realise the same the report list is being used in the add to report modal as well as the main report list page (but of course that makes good sense!) I'm not totally done from a design perspective but I wanted something more visually pleasing at least. I didn't know where to put the image files so I used base64 and included them as JS files. Seemed to make sense but maybe not the best approach in the future. Works well though! 😉 

### Review
Be great if either @federicoweber or @msanroman could take a look! Make sure what I have here is ok from a technical point of view. 👌 

#### Staging Deployment
https://task-improving-empty-states-analyze.dev.buffer.com :smile:_


moreofmorris:
Hey there @msanroman 👋 Thanks for moving the image to S3. Is there a way I can do this myself? I may want to still change those images. I totally agree with you on the comparison empty state but I found I couldn't find a suitable state that would reflect what I wanted. The profiles "get selected" as soon as they're checked in the dropdown and I couldn't find any other state that would allow me to only turn off the empty state once the user had truly selected them for comparison. Any help there would be great! ...I left it as it was, for now, as it was better than nothing.

PR #123, opened 2018-01-23
Add "Send Feedback" link to sidebar

tomredman:
### Purpose
Add a simple `mailto` feedback to the sidebar.
screen shot 2018-01-23 at 5 15 55 pm

#### Staging Deployment


packages/nav-sidebar/components/FeedbackLink.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Link
+} from '@bufferapp/components';
+
+const feedbackSpanStyle = {
```
tiggreen: We could use styled-components and replace this block with:

```
  const FeedbackSpan = styled.span`
     // all your styles come here. Plain CSS format.
  `;
```

Then, you could use it like this `<FeedbackSpan />` instead of `<span style={feedbackSpanStyle}>`. Cool huh? 😄 

We use styled-components so will be nice to follow the same pattern here.

packages/nav-sidebar/components/FeedbackLink.jsx
```diff
@@ -0,0 +1,35 @@
...
+	display: 'block',
+	padding: '0.75rem 0.5rem',
+	margin: '0px 0.5rem',
+	borderRadius: '4px'
+};
+
+const feedbackLinkStyle = {
```
tiggreen: Same here!

packages/nav-sidebar/components/FeedbackLink.jsx
tomredman: Awesome! Yes I'll definitely do this :)

PR #137, opened 2018-01-29
remove hourly chart

federicoweber:
### Purpose
To remove the hourly chart, as we are going to deprecate it with the GNIP changes.

#### Staging Deployment


federicoweber:
The coverage was showing a reduction of -0.06%; I've merged it anyway as it's not really relevant.

karamcnair:
Ooh, this is an interesting situation! Since we have no customers on at the moment and they won't have access to hourly charts to add them to reports in future (right?) this particular problem shouldn't occur but does this mean that either:

- any changes we make to charts in future will need to be backwards compatible because existing reports may still use them or
- we have to update user's existing reports if they use data or charts that we're deprecating?

msanroman:
Yeah @karamcnair! Federico and I were chatting around this yesterday on our technical 1:1 and we decided:

- Federico will make reports more resilient by ensuring that we only try to render elements that are existing (as well as performing RPC calls that are there, too). So a chart that's been removed from the codebase will simply disappear from a report, but shouldn't _break_ the report.
- Updating charts within a report can be DB expensive, but we _could_ update the codebase to auto-clean-up if during the "existence check" mentioned in the previous point we trigger a API call to delete a deprecated chart from a report. So that would be performed automatically on user request, one report at a time.

PR #141, opened 2018-02-02
Batch performance tracker RPC requests

federicoweber:
### Purpose
To make sure performance tracking is not interfering with network performances.

### Notes
**This sends out all the measures every 10 seconds**. I think this is a good balance between getting the data and making sure we are not interfering with the main RPC requests.

Given this interval, there is a chance we might lose some measures toward the end of the navigation, but for the cases, we are using the tracking at the moment I doubt this will be a concern, and we can always revisit the interval if needed :smile: .

#### Staging Deployment

### DIFF
--- a/packages/performance-tracking/middleware.js
+++ b/packages/performance-tracking/middleware.js
@@ -6,11 +6,20 @@ import {
 
 import { actionTypes, actions } from './reducer';
 
-export const storeMeasure = dispatch => (data) => {
+const pendingMeasures = [];
+let rpcCallTimeout;
+
+function sendMeasuresToRPC (dispatch) {
   dispatch(fetchActions.fetch({
     name: 'performance',
-    args: { data },
+    args: { measures: pendingMeasures.splice(0, pendingMeasures.length) },
   }));
+}
+
+export const storeMeasure = dispatch => (measure) => {
+  pendingMeasures.push(measure);
+  if (rpcCallTimeout) clearTimeout(rpcCallTimeout);
+  rpcCallTimeout = setTimeout(sendMeasuresToRPC, 10000, dispatch);
 };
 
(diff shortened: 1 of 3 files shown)


packages/performance-tracking/middleware.js
```diff
@@ -6,11 +6,20 @@ import {
...
 
-export const storeMeasure = dispatch => (data) => {
+const pendingMeasures = [];
+let rpcCallTimeout;
+
+function sendMeasuresToRPC (dispatch) {
   dispatch(fetchActions.fetch({
     name: 'performance',
-    args: { data },
+    args: { measures: pendingMeasures.splice(0, pendingMeasures.length) },
```
msanroman: This is a clever one haha 😄 - for the sake of code clarity, I'd still prefer this being `measures: pendingMeasures` and then resetting `pendingMeasures` to `[]` explicitly after the call, as somebody might not realize this `splice` call is emptying the array at the same time as it's returning all its values.

packages/performance-tracking/middleware.js
```diff
@@ -6,11 +6,20 @@ import {
...
   }));
+}
+
+export const storeMeasure = dispatch => (measure) => {
+  pendingMeasures.push(measure);
+  if (rpcCallTimeout) clearTimeout(rpcCallTimeout);
+  rpcCallTimeout = setTimeout(sendMeasuresToRPC, 10000, dispatch);
```
msanroman: Maybe we can move `10000` to be a constant like `TRACKING_DELAY`, or something like that?

tiggreen:
Amazing, this will also make the developer console network tab to be cleaner and easy to find the call! 👍

federicoweber:
Great points @msanroman will address those in a separate PR :smile:

PR #152, opened 2018-02-06
Comparisons Charts multi profiles tool-tip 

federicoweber:
### Purpose
To show the data for all the selected profiles in the Comparison Chart tooltip.
BAN-68

20180205_185805-screenshot

### Notes
Similar to what we are doing in the footer, we are now using the `profiles` collection to extrapolate the profiles data.

#### Staging Deployment

### DIFF
--- a/packages/audience-comparison-chart/mocks/dailyData.js
+++ b/packages/audience-comparison-chart/mocks/dailyData.js
@@ -4,7 +4,7 @@ export default [
       {
         day: 1504137600000,
         metric: {
-          label: 'Fans',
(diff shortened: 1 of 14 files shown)


federicoweber:
Codecov is pointing out that this line `<ChartTooltip metrics={this.points.map(p => p.point.metricData)} day={point.x} />,` is not fully covered. 
I'm going to merge this in, as i think that is acceptable, but I'm happy to rethink the testing strategy for that if we really want to have it covered :smile:

PR #154, opened 2018-02-06
Fix report rendering

msanroman:
### Purpose

To fix chart scales in report exports.


packages/compare-chart/components/Chart/chartConfig.jsx
```diff
@@ -20,7 +20,7 @@ export function truncateNumber() {
   return number;
 }
 
-export const highChartsConfigXAxis = {
+export const getXAxis = () => ({
```
federicoweber: :1st_place_medal:

PR #156, opened 2018-02-07
Better empty reports message

moreofmorris:
### Purpose
Make the empty state message a little nicer on the reports list page. This relates to this JIRA ticket: https://buffer.atlassian.net/browse/BAN-97

### Notes
Because the reports modal and the reports page both use the same reports list component I had to add a little prop that gets passed down from the modal to the list so it doesn't show the extra empty state text. That's because within modal that extra piece of text doesn't make much sense. I hope these sentences make sense! 😅  

### Review
Be great to get a quick look over this @msanroman! I'm trying to keep my pull request super small and simple these days! 😄 

#### Staging Deployment
https://http://task-better-empty-reports-message-analyze.dev.buffer.com/ :smile:_


moreofmorris:
Sounds great @msanroman, I'll do this as soon as possible! Makes so much sense! 👍

moreofmorris:
Hey there @msanroman, how do the new changes feel? :)

PR #157, opened 2018-02-07
Multi-Profile Picker: When 5 profiles are selected, we should dim/disable clicking on more than 5 profiles

tiggreen:
Here is the [JIRA card](https://buffer.atlassian.net/secure/RapidBoard.jspa?rapidView=17&modal=detail&selectedIssue=BAN-62) for more details.

#### Staging Deployment


tiggreen:
Deleted an unnecessary snapshot test hence the coverage decrease.

PR #166, opened 2018-02-09
When viewing a report, clicking on "Reports" should take you back to the Reports list

tiggreen:
[JIRA CARD. ](https://buffer.atlassian.net/secure/RapidBoard.jspa?rapidView=17&projectKey=BAN&modal=detail&selectedIssue=BAN-103)

### Review
Hey @federicoweber thanks so much for all the help you provided on this. Very accidentally I ended up chatting with Dan about this issue and he helped me find a different way without using the throttle. Also this way we separate the view part from the logic. 

#### Staging Deployment


federicoweber:
oh great one @tiggreen this is cleaner :100:

PR #182, opened 2018-02-23
Persist date for reports

msanroman:
### Purpose

To allow users to pick a date range for reports, and have it persisted so that they don't need to reselect them every time they want to check a report. The date range can be either static (e.g: Jan 1st - Jan 31st) or dynamic (e.g: Past 7 days).


packages/date-picker/reducer.js
```diff
@@ -65,21 +111,44 @@ export default (state = initialState, action) => {
...
+      return {
+        ...state,
+        loading: false,
+      };
     case `analytics_start_date_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
       return {
         ...state,
         loading: false,
         minDate: action.result,
+        presets: state.presets.map((preset) => {
```
federicoweber: Hey @msanroman, how do you feel about moving this to a dedicated function, and maybe breaking it down a bit?

Maybe something like this:
```
function enforceValidPresets(presets, validStartDate) {
	return presets.map((preset) => {
		const presetStartDay = moment().startOfDay('day').subtract(preset.range, 'days');
		preset.disabled = validStartDate > presetStartDay;
		return preset;
	}); 
}
```

packages/report/middleware.js
```diff
@@ -12,40 +13,69 @@ const getReportId = (pathname) => {
   return routeMatch ? routeMatch[1] : null;
 };
 const isReportDetailRoute = pathname => getReportId(pathname) !== null;
-const isExportRoute = pathname => pathname.match(/export\/reports/) !== null;
+const isRouteDifferentThanCurrentOne = (router, pathname) =>
```
federicoweber: :heart:

packages/report/middleware.js
```diff
@@ -12,40 +13,69 @@ const getReportId = (pathname) => {
...
-      if (isReportDetailRoute(action.payload.pathname) && !isExportRoute(action.payload.pathname)) {
+    }
+    case `update_report_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
+      if (!Object.is(state.report.dateRange, action.args.dateRange)) {
         store.dispatch(actions.fetch({
           name: 'get_report',
           args: {
-            _id: getReportId(action.payload.pathname),
+            _id: action.result._id,
+            timezone: moment.tz.guess(),
```
federicoweber: great call on using the user timezone in here :100:

packages/report/middleware.js
```diff
@@ -12,40 +13,69 @@ const getReportId = (pathname) => {
...
+          },
+        };
+        store.dispatch(reportActions.saveChanges(report));
+      }
       break;
-    case LOCATION_CHANGE:
-      if (isReportDetailRoute(action.payload.pathname) && !isExportRoute(action.payload.pathname)) {
+    }
+    case `update_report_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
+      if (!Object.is(state.report.dateRange, action.args.dateRange)) {
```
federicoweber: Hey @msanroman wondering if we can use `dateRangesDoNotMatch` in here too?

Asking because I like the added readability and because `Object.is({foo: 'foo'}, {foo: 'foo'})`, will return false, and that could happen if we change the way we pass the range to the action.

packages/report/middleware.test.js
```diff
@@ -17,6 +18,10 @@ describe('middleware', () => {
     },
     report: {
       id: 'report_id_1',
+      dateRange: {
+        startDate: '10/10/2016',
```
federicoweber: Super small one, I think we are using a unix timestamp format for the dates. Not an issue cause it will not influence this tests, but it could be confusing :smile:.

packages/server/rpc/getReport/index.js
```diff
@@ -13,8 +14,19 @@ const RPC_ENDPOINTS = {
   comparison: require('../comparison'), // eslint-disable-line global-require
 };
 
-function requestChartData (chart, startDate, endDate, session) {
+function calculateDateRange(range, timezone) {
```
federicoweber: thinking we might want to consider having a shared library between the front-end and the back-end for this manipulations. What do you think @msanroman?

PR #199, opened 2018-03-16
Fix toolbar height inconsistencies

msanroman:
### Purpose

Make toolbar height consistent by creating a Toolbar component.


moreofmorris:
Looks perfect @msanroman 🙌

PR #201, opened 2018-03-22
[BAN-125] Improve title of "Metrics Breakdown" when on Reports

msanroman:
### Purpose

Title for compare chart on a report is the selected metric label.


moreofmorris:
Looks great @msanroman 👍_...and it makes such a difference!_

PR #219, opened 2018-05-02
Fix/profile selector (ANA-204)

federicoweber:
### Purpose
To properly display long names in Profile selectors

#### Staging Deployment


federicoweber:
GOing to merge it in, I think the snapshot test coverage is ok as it is :smile:

PR #228, opened 2018-05-15
Report Logo Upload

tiggreen:
### Purpose

The goal of this PR is to implement the upload logo on Reports. Here is the styling guide to follow: https://paper.dropbox.com/doc/Custom-Logo-for-Reports-6GoJSHJ9JCAZNd5CuGQ8X

#### Staging Deployment


tiggreen:
Hey @federicoweber I'd love your review on this, in the meantime, I'll add tests to middleware to cover upload logo functionality with mocking. Thank you! 🌮

packages/report/middleware.test.js
```diff
@@ -291,5 +291,33 @@ describe('middleware', () => {
     });
   });
 
+  // it('UPLOAD_LOGO dispatches upload logo request', () => {
```
tiggreen: This is temporary and I'll add a test.

packages/report/index.test.jsx
```diff
@@ -93,7 +120,7 @@ describe('Report', () => {
     expect(component.props().moveUp('chart_id')).toEqual(actions.moveUp('chart_id'));
   });
 
-  it('move down should mispatch moveDown', () => {
+  it('move down should dispatch moveDown', () => {
```
federicoweber: :heart:

packages/report/reducer.js
```diff
@@ -94,6 +123,15 @@ export default (state = initialState, action) => {
         ...state,
         edit: true,
       };
+    case actionTypes.UPLOAD_LOGO:
+      return {
+        ...state,
+        logo: action.logo ? action.logo : state.logo,
+      };
+    case actionTypes.DELETE_LOGO:
```
federicoweber: Hey, @tiggreen I think this is not needed as it is, it's similar to the default handler :smile:. 

But :thinking: about it could be nice to unset the logo in here, and possibly leverage `delete_report_logo_${asyncDataFetchActionTypes.FETCH_FAIL}` to set it back in case of errors.

Reason for this is that the experience will be smoother as we will get immediate response for the action.

packages/report/reducer.test.js
```diff
@@ -60,6 +64,21 @@ describe('reducer', () => {
...
 
+  it('UPLOAD_LOGO sets uploading mode', () => {
+    const state = reducer(undefined, {
+      type: actionTypes.UPLOAD_LOGO,
+      logo: 'logo-image',
+    });
+    expect(state.logo).toEqual('logo-image');
+  });
+
+  it('DELETE_LOGO sets delete logo mode', () => {
```
federicoweber: Oh nice you already have a test for cleaning up the logo on `DELETE_LOGO` :100:, but I think this is not working as expected; the reason it's passing now is that the initial state is already empty.

I think that passing the initial state to the reducer, we will be able to test the removal:
```
  it('DELETE_LOGO sets delete logo mode', () => {
    const state = reducer({
      logoUrl: 'foo',
    }, {
      type: actionTypes.DELETE_LOGO,
    });
    expect(state.logoUrl).toEqual('');
  });
```

packages/report/reducer.test.js
```diff
@@ -191,6 +210,91 @@ describe('reducer', () => {
     });
   });
 
+  describe('uploading and deleting logo', () => {
```
federicoweber: Amazing work whit tests in here :heart_eyes:.

packages/report/reducer.js
tiggreen: Such a great point @federicoweber! I'll add `delete_report_logo_${asyncDataFetchActionTypes.FETCH_FAIL}` handing in the upcoming PR where I handle upload server error as well :)

packages/report/reducer.test.js
tiggreen: Oh right, that makes sense. Thanks for catching it, @federicoweber :) Will fix it now!

PR #235, opened 2018-06-08
[ANA-242] Improve how we handle Compare Chart legend values

tiggreen:
### Purpose
For more details please see this [JIRA issue.](https://buffer.atlassian.net/browse/ANA-242).

### Notes
This is only for Compare Chart because other charts/table using `GridItem` component don't have prevValue in the metrics. We can handle them in a different PR because it will require more and different type of work than this. cc @federicoweber 

#### Staging Deployment

### DIFF
--- a/packages/shared-components/TruncatedNumber/index.jsx
+++ b/packages/shared-components/TruncatedNumber/index.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import numeral from 'numeral';
 
-const TruncatedNumber = ({ children, absoluteValue, shorterOption }) => {
+const TruncatedNumber = ({ children, absoluteValue, shorterOption, showPercentSign }) => {
   let number = children;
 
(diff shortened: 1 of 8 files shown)


tiggreen:
I'm making some changes to this PR can you please hold for the review? I'll update you when it's good to go for review!

packages/shared-components/TruncatedNumber/index.jsx
```diff
@@ -17,25 +17,31 @@ const TruncatedNumber = ({ children, absoluteValue, shorterOption }) => {
     formattedNumber = numeral(number).format('0a');
   } else if (number >= 10000) {
     formattedNumber = numeral(number).format('0.0a');
+  } else if (number < 0.1 && number > 0) {
```
federicoweber: I think we can remove this one and just use `(number < 1 && number > 0 && showPercentSign)`. That will get us covered for all % < 1 right?

PR #256, opened 2018-06-25
Add cover page for reports when exporting

moreofmorris:
### Purpose
This pull request is just for adding the cover page for reports when exporting. It doesn't have all the other report features (like consistent header and footer etc). I've tried quite a few different variations for the cover page and it seems the centralised title and date works best for reports with images. I'm sure there is more we can do but it could be good to get it out there and see what people say. Though I'm aware it's not the most exciting cover page right now! Maybe we can do things like colour tinting and such in the near future.

### Review
Be great to get your eyes on this @federicoweber 👍 As for the review, on speaking with Mike he mentioned that we shouldn't worry too much when the test coverage is failing on `story.jsx` files. Aside from this, the code should be pretty simple. The design is stripped back at the mo, we can just put it out there and see what customers say and build from there.

#### Staging Deployment
https://task-cover-page-for-reports-analyze.dev.buffer.com/ :smile:_

### DIFF
--- a/packages/compare-chart/__snapshots__/snapshot.test.js.snap
+++ b/packages/compare-chart/__snapshots__/snapshot.test.js.snap
@@ -2071,7 +2071,7 @@ exports[`Snapshots CompareChart should render the compare chart 1`] = `
           <div
             style={
               Object {
-                "margin": "1.5rem 1.5rem 1rem",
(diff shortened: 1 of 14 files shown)


moreofmorris:
I think I agree with you @federicoweber on the redundant header. I will try some things out today to see what works 👍

PR #257, opened 2018-06-25
fix page break for posts table

federicoweber:
### Purpose
To fix reports pagination for Posts Tables

### Notes
This was a fun bug to track down :smile: .

The new layout was breaking the page break calculations and that was an easy fix, but then I realized that even though the calculations were working fine, the pdf was ignoring the `page-break-before` rules. Turned out rendering to pdf is ignoring page breaks for `table`; therefore I've changed the markup a bit and we are now using ordered lists.

I've added in a test to make sure the page break logic is working fine with this new layout, but that is still a bit brittle, as it's testing against a mocked layout. :thinking: It could be worth exploring some integrations tests for the export, as it's quite coupled with the actual components layouts and i think this is not the first time this is breaking cause of a layout change.

Also tok the chance to update the style of the page break border :straight_ruler:.

### Review

Hey @msanroman, would love your eyes on this one, as you are the most familiar with the PDF export logics :smile:.

@moreofmorris I'm tagging you in too, to keep you in the loop on this one :smile:. 

#### Staging Deployment

### DIFF
--- a/packages/posts-table/__snapshots__/snapshot.test.js.snap
+++ b/packages/posts-table/__snapshots__/snapshot.test.js.snap
@@ -345,11 +345,11 @@ exports[`Snapshots PostItem should not render the top posts item when no metrics
       }
     }
   >
-    <tr
-      className="sc-hSdWYo kccNLr"
+    <li
+      className="sc-hSdWYo kVofMF"
     >
-      <td
-        className="sc-cvbbAY ePXgRF"
+      <div
+        className="sc-cvbbAY brUDDv"
       >
         <span
           style={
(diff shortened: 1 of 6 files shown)


moreofmorris:
Hey there @federicoweber and @msanroman 👋

Just a heads up, Mike previously had the Posts Breakdown table with <div>s but when it came to introducing the side numbers I wanted to ensure that the columns wouldn't change widths per row based on how big the numbers we (1 would be one column width, 10 would be another column width, 100 would be another column width) and so I resorted to using <table> to ensure columns behaved themselves all the way down. I think that must of have the side effect of page breaks no longer working as they should. I apologise if that was what caused this bug! 😓 

I'm working on designs at the moment that will change fundamentally how we display Post Breakdowns data in reports. But it's fantastic to see the page break situation fixed like this, really appreciate it! And sorry again if it was me that caused extra work here. As long as we don't run into a situation where columns start to shift around based on how large the side numbers are, then we're all good 👌 

Looks good to me @federicoweber! Thank you! 👍

federicoweber:
Thanks for your review @moreofmorris. 
Very good point on the side numbers, I went with a fixed width for that, but keen to review that if you feel it's needed.

And no worries at all, we should have caught that with tests.

PR #262, opened 2018-07-11
[ANA-24] Reports should have an optional editable description

tiggreen:
### Purpose
Here is the [JIRA card](https://buffer.atlassian.net/browse/ANA-24) for more context.

#### Staging Deployment

### DIFF
--- a/packages/report/__snapshots__/snapshot.test.js.snap
+++ b/packages/report/__snapshots__/snapshot.test.js.snap
@@ -543,7 +543,7 @@ exports[`Snapshots ChartEditButtons they render 1`] = `
 exports[`Snapshots Cover renders a cover with a logo 1`] = `
 <span>
   <section
-    className="sc-dnqmqq gjnmZE"
+    className="sc-iwsKbI hgkCET"
   >
     <article
       className="sc-bxivhb cEKEIA"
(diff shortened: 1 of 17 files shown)


tiggreen:
Hey @moreofmorris! You can test the report description from the design perspective https://task-report-description-analyze.dev.buffer.com/. Would love to know what you think :)

moreofmorris:
Hey there @tiggreen! First things first, wow, this came together super quickly! Really appreciate this! It's been great to actually play with this and see how it works in the wild. A few things have come to mind... I know time is short, so I'll separate them out to what I see as essential and nice-to-haves.

**Essential**
- On the report page make the description text `font-weight 400`, `line-height: 1.4rem`, `font-size: 1rem`. I'm basing that off some changes I did in the dev tools, reading the computed styles. I know it doesn't quite translate to our Text component but something near to that would be great. _(See the screenshot)_
- Move the hover icons for the description (and the title, so to match) to the left of the input boxes when hovering. When the description is multiple lines, the pen icon looks kinda awkward. I'd imagine most descriptions to be multiple lines. _(See the screenshot)_
- Cap the maximum amount of characters for the description to around 300 characters. Something like 3 lines of text-ish. This might be too harsh, but I wouldn't want the other way either where it's reams of text for something that's just meant to be a little note on the report.

**Nice-to-haves**
- Could the input be a text area instead? That could feel nicer, especially as the description leans to multiple lines. Not the end of the world though! :)

screenshot 2018-07-13 12 06 30
screenshot_2018-07-13_11_49_40

I think the output PDF looks good. I may make some changes in the near future but I'd love to try it out more first and see it being used. I don't think we should hold it up anymore, looks good!

Thanks again Tigran!

tiggreen:
hey @moreofmorris I've addressed all the changes you requested above including the textarea! :)

moreofmorris:
Hey there @tiggreen! After having those discussions with @federicoweber on Friday and then playing with it myself, it does make sense to revert this back to a simple text box. I'll address further design enhancements to the UX of many aspects of the report editing experience at a later date. In this state, it feels strange to not have a submit button and it feels strange that a user can enter line breaks. The report description should only be a quick note. Once we revert the textarea back to a single line text box (just like the report title) we're good to go! Thanks so much. And thanks for being patient as we went back and forth a bit 👍

PR #263, opened 2018-07-17
Report cover updates

moreofmorris:
### Purpose
This contains a few visual updates to the reports cover page. Things were looking a little hectic and it needed calming down a bit visually. I've rebalanced a few of the weights and sizes in the reports and in the chart headers so everything feels a little more consistent from the graphs to the reports to the PDF export.

### Notes
I was able to remove some code that didn't seem needed anymore as the chart headers for app and reports is now the same :)

### Staging Deployment
This repo is CI/CD enabled and staging deployment should be available at:
https://better-report-cover-analyze.dev.buffer.com/ :smile:

### DIFF
--- a/packages/report/components/Cover/index.jsx
+++ b/packages/report/components/Cover/index.jsx
@@ -41,9 +41,9 @@ const Inner = styled.div`
 const Title = styled.h1`
   display: block;
   color: ${black};
-  font-size: 2.2rem;
+  font-size: 2.4rem;
   font-weight: 500;
-  margin: 0 0 1rem;
+  margin: 0 0 0.8rem;
 `;
 
 const Description = styled.h2`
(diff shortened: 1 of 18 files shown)


moreofmorris:
Thanks for looking this over @federicoweber 👍

packages/report/components/Cover/index.jsx
```diff
@@ -15,73 +15,86 @@ const Page = styled.article`
...
     <Container>
-      <Wrapper>
+      <div>
         <Inner>
           <Text>
             <Title>{name}</Title>
-            <DateRange {...dateRange} large />
-            <Description>{description}</Description>
+            <DateRange {...dateRange} />
+            {description && <Description>{description}</Description>}
```
federicoweber: :100:

PR #264, opened 2018-07-24
Add Service Worker for rpc requests

federicoweber:
### Purpose
To speed up Analyze loading by caching RPC requests on the client side

### Notes
This introduces a `@bufferapp/RPC-worker` component, that is a Service Worker specifically designed to handle our RPC endpoints.

As we want to be specific with the RPC requests we are caching (no action or performance tracking for example), we are **using a whitelist approach**. You can specify that when you are instantiating the worker passing down the name of the RPC endpoints to match and the strategy to use
```
import rpcWorker, { STRATEGIES } from '@bufferapp/RPC-worker';

rpcWorker('0.0.1', [
  { name: 'environment', strategy: STRATEGIES.cacheFirst },
  { name: 'user', strategy: STRATEGIES.cacheFirst },
  { name: 'profiles', strategy: STRATEGIES.cacheFirst },
  { name: 'profiles_overview', strategy: STRATEGIES.cacheFirst },
]);
```
_As of now there is only one predefined strategy `cacheFirst`, but it's also possible to pass a custom method, and more will follow._

The first parameter provided is a **version number**, that is used to construct the caching namespace, **so if you want to invalidate, and delete, old cache just bump that up**.

#### Couple of thing to consider before adding it in:
- As of now this is caching new data for the same request each day, if the dates changes in the request, we need to define a strategy to invalidate old data, or it will pile up.
- with Cache First strategy we are getting the data again if we are not using it. maybe we can send that back in the browser if the result is different from the cached one. As getting last data will require a reload.
- Could be interesting to add a strategy that can force data invalidation after midnight (browser timezone)

### Review
Hey @hharnisc and @msanroman, this is still a WIP, and will add better testing for the caching strategies before merging it in, but would love to pick your brains  around the general flow :smile:.
Thanks!

This is not quite working at the moment in the dev environment cause it has some problems with webpack server.

#### Staging Deployment


packages/rpc-worker/index.js
```diff
@@ -0,0 +1,117 @@
+export const STRATEGIES = {
+  cacheFirst(request) {
+    return new Promise((resolve, reject) => {
+      request.clone().json()
```
hharnisc: I wonder if we can iron this out with async/await :thinking:

federicoweber:
Hey @hharnisc, good point on considering caching at the redux level.
My reasons for going with service workers is that they are quite self-contained, easy to invalidate ant reason with, and could be extended to cache the rest of the app: script assets, …, to fully accive offline functionality.

Do you have any specific concern in mind regarding this approach and/or porting it approach over to Publish and Account?

packages/rpc-worker/index.js
federicoweber: I clean this up a bith moving out the two strategies.

tomredman:
@hharnisc quick ping to keep this discussion alive :)

@federicoweber I've spotted a couple oddities. The main one is I catch the Overview reloading a second time: http://hi.buffer.com/4207fa5d5360 (just watch the first ~10 seconds or so, I tried to reproduce it but couldn't do it consistently). 

I was able to reproduce this fairly frequently after clicking around.

Another thing I saw were some numbers weren't for Brian. P's account briefly flash on the Overview. I'll try to get a video of that one too.

federicoweber:
Hey @tomredman thanks for giving this a run. Will look into the double reload, and the mysterious data (on this one I think this is something can happen in production version as well if we switch fast between profiles). 

I think most of the issues with quick page / profiles switching are due to the fact that we are requesting the data on every change, and not properly invalidating old requests, this is something we need to address in general.

Curious in term of feel and experience, doing regular navigation (no stress test :smile:), how does this feel to you?

tomredman:
Navigating around normally feels great @federicoweber

federicoweber:
Going to ship this :shipit:

PR #267, opened 2018-08-01
Update README.md

hharnisc:
#### Staging Deployment


msanroman:
minimum-review-bot does not agree with my comment being good enough to be a review 😅

hharnisc:
@msanroman looks like it took a second but it went through 😀

PR #271, opened 2018-08-06
List reports by organization

msanroman:
### Notes
To be merged after the migration.

### DIFF
--- a/packages/report-list/middleware.js
+++ b/packages/report-list/middleware.js
@@ -4,11 +4,11 @@ import { actionTypes as reportListActionTypes } from './reducer';
 
 export default store => next => (action) => { // eslint-disable-line no-unused-vars
   switch (action.type) {
-    case `user_${actionTypes.FETCH_SUCCESS}`:
+    case `profiles_${actionTypes.FETCH_SUCCESS}`:
       store.dispatch(actions.fetch({
         name: 'list_reports',
         args: {
-          userId: action.result.id,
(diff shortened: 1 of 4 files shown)


packages/report-list/middleware.js
```diff
@@ -4,11 +4,11 @@ import { actionTypes as reportListActionTypes } from './reducer';
 
 export default store => next => (action) => { // eslint-disable-line no-unused-vars
   switch (action.type) {
-    case `user_${actionTypes.FETCH_SUCCESS}`:
+    case `profiles_${actionTypes.FETCH_SUCCESS}`:
       store.dispatch(actions.fetch({
         name: 'list_reports',
         args: {
-          userId: action.result.id,
+          organizationId: action.result[0].organizationId,
```
federicoweber: Hey @msanroman was wondering: why we are wrapping the result in an array? Do you think we can directly return the result, given that we are consuming only the first element anyway?

hharnisc:
Sorry for the noise on this PR -- the review bot wasn't configured to receive the webhook for submitted reviews. That's all fixed up now :ok_hand:

packages/report-list/middleware.js
msanroman: It's grabbing the `organizationId` from the first profile of the list. We're consuming the full profile list in the profile selector reducer 😛

federicoweber: Oh right :man_facepalming: :smile:

PR #273, opened 2018-08-13
Trial status on sidebar

msanroman:
## Purpose

To provide the trial status on the nav sidebar for Analyze. The status is displayed only for owners right now, as that's data that's only available for them. Once I merge and deploy https://github.com/bufferapp/buffer-web/pull/15382, it will become available for all team members too!

## Note

There is also a piece of code that wraps the trial status on a `Link` component that will become available only for the owner. For now, that's hardcoded to `false` in the profiles reducer, but I will create another PR that will return that information from the profiles endpoint on the API as well (whether the user requesting the profiles for the organization is the owner or not), and will tweak the reducer & RPC accordingly :)

### DIFF
--- a/packages/account/middleware.js
+++ b/packages/account/middleware.js
@@ -1,0 +1,17 @@
+import { actions, actionTypes as fetchActions } from '@bufferapp/async-data-fetch';
+
+export default store => next => (action) => { // eslint-disable-line no-unused-vars
+  switch (action.type) {
+    case `profiles_${fetchActions.FETCH_SUCCESS}`:
+      store.dispatch(actions.fetch({
+        name: 'get_account_details',
+        args: {
+          organizationId: action.result[0].organizationId,
+        },
+      }));
+      break;
+    default:
+      break;
+  }
+  return next(action);
+};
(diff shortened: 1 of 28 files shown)


packages/account/middleware.js
```diff
@@ -0,0 +1,17 @@
+import { actions, actionTypes as fetchActions } from '@bufferapp/async-data-fetch';
+
+export default store => next => (action) => { // eslint-disable-line no-unused-vars
+  switch (action.type) {
+    case `profiles_${fetchActions.FETCH_SUCCESS}`:
+      store.dispatch(actions.fetch({
+        name: 'get_account_details',
+        args: {
+          organizationId: action.result[0].organizationId,
```
federicoweber: This is working great but, it made me think if it makes sense to perhaps consider storing this in the user session :thinking:.

Not something we need for this PR, just wanted to leave some food for thoughts :smile:.

msanroman:
Yeah @federicoweber! Thanks!

I've squashed the trial status work on one commit but left two other commits outside of that since they aimed at different goals than the main chunk of work 😄

PR #274, opened 2018-08-13
Proxy reports stored on S3

msanroman:
## Purpose

To allow reports bigger than 6MB to be downloaded as PDF too.

## Notes

I'll switch to the non `Dev` branch of the Lambda function once I've uploaded its new implementation there - doing it in two phases to ensure there is no downtime on downloading reports.

### DIFF
--- a/packages/posts-table/components/PostsTable/components/PostItem/index.jsx
+++ b/packages/posts-table/components/PostsTable/components/PostItem/index.jsx
@@ -145,7 +145,7 @@ const ContentBox = styled.div`
 
 const Attachment = ({ type, media }) => {
   if (['picture', 'photo', 'video', 'image'].includes(type) && media) {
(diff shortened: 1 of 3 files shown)


packages/posts-table/components/PostsTable/components/PostItem/index.jsx
```diff
@@ -145,7 +145,7 @@ const ContentBox = styled.div`
 
 const Attachment = ({ type, media }) => {
   if (['picture', 'photo', 'video', 'image'].includes(type) && media) {
-    return <MediaThumbnail alt="" crossOrigin="Anonymous" src={`https://safeimage.buffer.com/${media.thumbnail}`} />;
+    return <MediaThumbnail alt="" crossOrigin="Anonymous" src={`https://safeimage.buffer.com/224x/${media.thumbnail}`} />;
```
federicoweber: :heart: great one @msanroman.

PR #276, opened 2018-08-13
[ANA-312] Tweak analyze early access to be based on enabled profiles

tiggreen:
### Purpose
Instead of checking to see if the user has `analyze_early_access` feature flip we can simply check if the user has any analyze enabled profiles or not. 

### Notes
This PR will be merged after we ran the mongo script that will disable all analyze active profiles that don't have analyze subscriptions anymore. 

#### Staging Deployment


tiggreen:
I'm going to merge this only after @msanroman 's PR is merged :)

PR #284, opened 2018-09-04
Alphabetically sort profiles in the selector

federicoweber:
### Purpose
To properly sort profiles in the selectors

#### Staging Deployment

### DIFF
--- a/packages/server/rpc/profiles/index.js
+++ b/packages/server/rpc/profiles/index.js
@@ -2,6 +2,15 @@ const { method } = require('@bufferapp/micro-rpc');
 const rp = require('request-promise');
 const profileParser = require('../utils/profileParser');
 
+function getComparableUsername(username) {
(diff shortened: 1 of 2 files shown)


packages/server/rpc/profiles/index.js
```diff
@@ -2,18 +2,27 @@ const { method } = require('@bufferapp/micro-rpc');
 const rp = require('request-promise');
 const profileParser = require('../utils/profileParser');
 
+function getComparableUsername(username) {
+  return username.replace('@', '').toUpperCase();
```
tiggreen: curious why do we have to convert it to uppercase here?

federicoweber: great one @tiggreen reason for this is that
```
'foo'.localeCompare('Foo') = -1
'foo'.localeCompare('foo') = 0
```

Not a big one, but I tend to prefer flat case comparison :smile:.

PR #285, opened 2018-09-10
Add Hashtags table in posts tab (ANA-318)

federicoweber:
### Purpose
To add the hashtags table in the posts tab

### Notes
This also updates the style of the post table rank number so that it matches with the hashtags table.

#### Staging Deployment


tiggreen:
Amazing work @federicoweber and It works great and quite fast. 🎉 

Hey @federicoweber not sure just yet if this PR is WIP or not but while I was reviewing it I noticed two small things. 

1. When no hashtag is found dates are coming as invalid: http://hi.buffer.com/e49a74bbbc30.
2. Is there any reason why the numbers in the table are right aligned but everything else in the table is left aligned? http://hi.buffer.com/a5b4bdff2009

federicoweber:
Hey @tiggreen thanks for looking at this one. Still WIP, but great catch on the invalid dates will fix it soon :smile: . As for the rank alignment that is by design specs.

packages/shared-components/ArrowIcon/index.jsx
```diff
@@ -11,31 +11,34 @@ const Container = styled.div`
   }
 `;
 
-const ArrowIcon = ({ diff }) => {
-  if (diff > 0) {
+const ArrowIcon = ({ diff, upward, downward, color }) => {
```
tiggreen: Awesome stuff on moving this component into `shared-components`. 🎉 ❤️

packages/hashtags-table/components/TableRows/index.jsx
```diff
@@ -0,0 +1,258 @@
...
+  box-sizing: border-box;
+  position: relative;
+  color: ${outerSpace};
+  padding: 1.1rem 1rem 1.25rem 1rem;
+  text-align: left;
+  vertical-align: top;
+  flex-shrink: 0;
+`;
+
+const NumberCell = styled.div`
```
msanroman: Could we get rid of all the duplication by having a `Cell` styled component with the generic styles, and have the rest extend from it? Like: ```const NumberCell = styled(Cell)`<-- styles for NumberCell -->`;```

packages/hashtags-table/middleware.js
```diff
@@ -0,0 +1,62 @@
...
+import { actionTypes as profileActionTypes } from '@bufferapp/analyze-profile-selector';
+import { actionTypes as dateActionTypes } from '@bufferapp/analyze-date-picker';
+import { actionTypes as exportActionTypes, actions as exportActions } from '@bufferapp/analyze-png-export';
+import { actionTypes as exportCSVActionTypes, actions as exportCSVActions } from '@bufferapp/analyze-csv-export';
+
+const formatDataForCSVExport = ({ metrics }) => {
+  const data = { hashtags: [] };
+  metrics.forEach((hashtag) => {
+    const hashtagName = hashtag.display_name;
+    delete hashtag.display_name;
```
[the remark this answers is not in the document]
msanroman: +1, not sure why this line :)

packages/server/rpc/hashtags/index.js
```diff
@@ -0,0 +1,87 @@
...
+    return requestHashtags(
+      profileId,
+      profileService,
+      dateRange,
+      req.session.analyze.accessToken,
+      req.app.get('analyzeApiAddr'),
+    )
+      .then(({ response }) => {
+        return {
+          hashtags: extractMetrics(response.hashtags),
```
msanroman: Why do we need to apply a data transformation? I'd love for us to simplify as much as possible each step in our API calls, so if we need to tweak how we're responding with these new endpoints, we can do so now :)

packages/hashtags-table/middleware.js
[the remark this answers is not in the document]
federicoweber: it was to remove it from the following loop, but I ended up implementing a simpler strategy. Thanks for pointing it out :smile:.

packages/server/rpc/hashtags/index.js
federicoweber: Hey @msanroman great one.

I totally agree with you on simplifying the RPCs as much as possible, but in this case, I think we want these transformations to happen in here. 

The reason is that we are preparing the data for the specific needs of the hashtag table rendering: only rendering the required metrics and passing down the labels we want. While from the APIs we are returning the full list of metrics for each hashtag. I think this is a reasonable approach as it keeps the APIs free from the knowledge of how the data is displayed.

Another option could be to move this directly on the Hashtag table itself. I think it might be worth to move it there if we start to use hashtags in multiple visualizations.

Looking forward to chat more about this :smile:.

PR #286, opened 2018-09-11
micro-rpc -> buffer-rpc

hharnisc:
### Purpose

Uses the Buffer RPC handler, which removes the dependency on micro while still provide async/await handlers in an express server

### Notes

This is backwards compatible with Micro RPC

### Review

Letting this run for a bit on analyze to make sure there aren't any longer running issues.

#### Staging Deployment


hharnisc:
Great question @msanroman :raised_hands: 

The big thing is that to make `micro-rpc` work with express we have to wrap it with a `then/catch` to handle async:

https://github.com/bufferapp/buffer-analyze/pull/286/files#diff-068e7c6acbb89d55e9fa2192f813a6d8L107

`micro-rpc` also reads the request body in a way where it can't be read again (easily) by other middlewares/handlers.

```
import { json } from 'micro'

export (req, res) => {
  const body = json(req) // this reads from a stream, all the way to the end
}
```

It is equivalent to this problem (reading a stream twice): https://stackoverflow.com/questions/19553837/node-js-piping-the-same-readable-stream-into-multiple-writable-targets/19561718

The express way to solve this problem (quick and dirty), is to read the stream and set on on the request object with `body-parser`.

Since we've chosen to use express since it is more ubiquitous (and more errors googleable :joy:) `buffer-rpc` just does things the express way -- including calling `next` under the hood so it can be used just like any other express handler.

TL;DR this makes things work consistently with what you'd expect from an express request handler.

hharnisc:
Another thing to note @msanroman -- in the next few days/weeks I plan on migrating the client as well. The only new feature of `buffer-rpc` is that it exposes a `x-buffer-rpc-name` header that can be used instead of the `name` in the json body. This allows us to easily get the rpc that is being called in `datadog`, Logs and even do things like routing traffic based off headers if we needed to (splitting up a service for example).

PR #287, opened 2018-09-12
Filter posts

msanroman:
## Purpose

To be able to filter the post breakdown table by different keywords, and store that search in a report.

## Notes

Even though codecov is complaining, I think we're good as the number of lines in the codebase has increased and the coverage diff is just -0.02% :)

### DIFF
--- a/packages/posts-table/components/PostsTable/components/BreakdownLegend/sortingDictionary.js
+++ b/packages/posts-table/components/PostsTable/components/BreakdownLegend/sortingDictionary.js
@@ -5,7 +5,7 @@ export default {
   },
   post_impressions: {
     ascending: 'most',
-    descending: 'fewest'
+    descending: 'fewest',
   },
   post_reach: {
     ascending: 'highest',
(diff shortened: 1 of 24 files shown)


packages/posts-table/components/PostsTable/components/BreakdownLegend/sortingDictionary.js
```diff
@@ -0,0 +1,59 @@
+export default {
```
federicoweber: Nothing do to right now, but I'm thinking it could be interesting to consolidate all our dictionaries into a shared module. We have many of those in a few different places.

packages/server/rpc/posts/index.js
```diff
@@ -2,29 +2,7 @@ const { method } = require('@bufferapp/micro-rpc');
 const rp = require('request-promise');
 const DateRange = require('../utils/DateRange');
 
-const mergeStatsWithUpdates = (stats, updates) => {
```
federicoweber: So nice to remove all of that :heart_eyes:

PR #305, opened 2018-11-13
Upgrade to Webpack 4 and add bundle splitting

msanroman:
### Purpose

- Upgrade to Webpack 4 and split code between bundle and vendor.
- Uses React production build (this clears dev warnings as well as serving an optimised build for production).
- Reduces bundle size by 45%.


packages/server/webpack.config.prod.js
```diff
@@ -0,0 +1,14 @@
+const merge = require('webpack-merge');
+// const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
+const common = require('./webpack.config.common.js');
+
+const merged = merge(common, {
+  mode: 'production',
+  devtool: undefined,
+  optimization: {
+    nodeEnv: 'production',
```
federicoweber: great one @msanroman

PR #306, opened 2018-11-14
[WIP] Use jest from root directory to run all unit tests and lerna snapshots in parallel to check snapshots for every package

msanroman:


gmzjuliana:
hey @msanroman do you think we can close this one?

msanroman:
> hey @msanroman do you think we can close this one?

Yesss! Great catch @gmzjuliana! Thanks! 🌮

PR #313, opened 2018-11-23
Fix/demographic render failures

msanroman:
### Purpose

To prevent this from happening:

analyze buffer com_audience_5ae0ca99152710256e8a7b95

Instead, we'll display the `NoData` component.


federicoweber:
Thanks @msanroman

PR #316, opened 2018-12-10
Tweak how we report best frequency to post

msanroman:
## Purpose

To have a nicer way to share the best frequency to post answer: `Once daily`, `Twice daily`, `Three times daily`, `Four times daily`, ... `10 times daily` etc. 😄


packages/answers-chart/components/Chart/index.jsx
```diff
@@ -2,28 +2,13 @@ import React from 'react';
...
-      break;
-    case 'frequency':
-      formatter = function () {
-        return `${this.value} daily`;
-      };
-      break;
-    default:
-      break;
-  }
+  const formatter = function () {
```
federicoweber: This is so much nicer :heart_eyes:.

PR #317, opened 2018-12-10
add Server Timing headers to rpc/compare (AT-35)

federicoweber:
### Purpose
To get a better sense of what is the bottleneck for Instagram API performances issues.

### Notes
We can examine those in Chrome network tab

20181210_120346-screenshot

#### Staging Deployment


gmzjuliana:
hey @federicoweber do you think we could close this one?

PR #337, opened 2019-01-18
Task/add rpc fail tracking

federicoweber:
### Purpose
To log failed RPC requests on DataDog and Bugsnag

### Notes
Could be worth passing down the error directly in the response. If we want to do that we should edit all the responses to return a JSON.

#### Staging Deployment


packages/server/rpc/utils/rpcErrorHandler.js
```diff
@@ -0,0 +1,38 @@
...
+const dogstatsd = new StatsD('dd-agent.default');
+
+function formatError(error, rpcName) {
+  return {
+    name: 'buffer.error.rpc',
+    tags: ['app:analyze', 'rpc', rpcName],
+  };
+}
+
+function logToDD(rpcName, error) {
```
tiggreen: Why not `logToDatadog`? :)

federicoweber:
Those are good point @msanroman. I initially thought to remove the `catch` statement and handle it all in the existing errors logging middleware, but that would have required adding error handling on the front-end, and I wanted to keep this PR focused on the error logging.

Having said so, you are right: surfacing errors to the front-end is something we will need to do sooner or later, and this could be the first step in that direction. Will give it a go 😉

federicoweber:
Hey @msanroman 👋, I was working on this one and noticed that `async_data_fecth` is triggering `FETCH_SUCESS` also in case of errors. I've added a PR there to address it https://github.com/bufferapp/buffer-js-async-data-fetch/pull/1. But will need to merge that in before we can continue the work on this one, otherwise we will need lot's of protective code in our `FETCH_SUCESS` handlers to check for properly formatted response data.

Good news is that we already have an Error state we can use, as we can use our [`ErrorBoundary`](https://github.com/bufferapp/buffer-analyze/blob/d3621e6ec68df5ba23d743ea3bf2063a688b7887/packages/shared-components/ErrorBoundary/index.jsx) for that, we just need to surface the error on the state on `FETCH_FAIL` 😉 .

msanroman:
Hey @federicoweber - can you send me an example or some steps to reproduce when this is triggering a `FETCH_SUCCESS` on errors? 😄

Would love to understand that better because in the past we were just catching exceptions that were thrown in our methods, and those unhandled exceptions should get into the `catch` block of the RPC middleware rather than the `then` one 🤔

federicoweber:
Hey @msanroman, sure thing you can play with it in `/test/rpc-error-handling` I've made the `summary` rpc throw an error.

a6165b74-9023-46c0-9905-a9967b0465f7 png
As you can see the `FETCH_SUCESS` is firing before the `FETCH_FAIL`, that the Promise will always process `then` before `catch`.
It is still possible to listen for `FATCH_FAIL` and maybe reset the state, but there will be cases were the reducer will still crash, as we are not protecting for malformed results, and I think we should not, I think we should not trigger the `SUCCESS` at all in case of an error.

msanroman:
Woooow! Great catch @federicoweber!

Sounds like an amazing plan! 👍

PR #346, opened 2019-02-12
Increase CPU limits

msanroman:
### Purpose

CPU is getting throttled in some pods. This should help to maintain
performance in occasional surges.

Thanks to @colinscape for spotting it! 🙏


tiggreen:
Thanks @msanroman, may I know what made you think to increase the CPU limit? :D

msanroman:
Sure @tiggreen! @colinscape mentioned the CPU was getting throttled in some pods, and the [Publish team dealt with it in the same way](https://github.com/bufferapp/buffer-publish/commit/08feef51b415adafa5b93fb8d3e8062cb7c89863) (they were having more issues, so this is more like a preventive thing on our end) 😄

PR #347, opened 2019-02-15
Fix missing days in compare chart (AT-56)

federicoweber:
### Purpose
To fix missing days in Compare chart when we had only partial data for the previous period 

### Notes
Hey @tomredman and @jameofmorris, as the chart was looking a bit jenky with partial data for the previous period, I took the liberty to disable the previous period toggle in case of partial data, happy to revisit if you prefer a different approach 😄.

image

#### Staging Deployment


tomredman:
Thanks for the ping @federicoweber :) This feels good to me. Thanks for thinking about this one!

tomredman:
Will test this as soon bundling is working again. I want to check it against the user in the ticket.

PR #351, opened 2019-02-25
Stories breakdown

msanroman:
## Purpose

To provide an Instagram Stories breakdown.


package.json
```diff
@@ -97,6 +97,7 @@
     "global": "^4.3.2",
     "react-dropzone": "^4.2.9",
     "react-loadable": "^5.5.0",
-    "redux-dynamic-middlewares": "^1.0.0"
+    "redux-dynamic-middlewares": "^1.0.0",
+    "styled-components": "4.1.1"
```
federicoweber: Hey @msanroman on the PR for the new nav I went with version 3.4.10, as from 4.0 on Styled components is using `React.forwardRef`, and that was breaking a few things around (this is also the cause of the failing snapshots tests on this PR 😉 ).
I think it's worth upgrading to 4.1, but was thinking to do that in a dedicated PR. What's your hunch on this one?

packages/server/rpc/stories/index.test.js
```diff
@@ -0,0 +1,40 @@
...
+    expect(stories.name)
+      .toBe('stories_breakdown');
+  });
+
+  it('should have the expected docs', () => {
+    expect(stories.docs)
+      .toBe('get a full breakdown of Instagram stories');
+  });
+
+  it('should send a POST request to /move_chart with the provided parameters', async () => {
```
federicoweber: I think the title in here is a bit off 😄

package.json
msanroman: Wow, great catch @federicoweber! 🙏 

Will do, and yeah - great point on the upgrade to 4.1 in a separate branch 😄

packages/server/rpc/stories/index.test.js
msanroman: 😅

PR #355, opened 2019-02-28
Upgrade styled components to 4.1.1

federicoweber:
### Purpose
To upgrade Styled components to `4.1.1` and fix some style issues we are seeing caused by multiple versions.

### Notes
- this is not upgrading to redux 6.0 and connected-react-router 6.0 because of this change that will break all our tests. _(will take care of that in a dedicated PR, as for now just wanted to fix the style issues we are seeing)_ 
> Passing store as a prop to a connected component is no longer supported. Instead, you may pass a custom context={MyContext} prop to both <Provider> and <ConnectedComponent>. You may also pass {context : MyContext} as an option to connect. [(ref.)](https://github.com/reduxjs/react-redux/releases/tag/v6.0.0-beta.1)

- Can't fully remove `react-redux-route` from dependencies as it's needed for `@bufferapp/app-sidebar`.

#### Staging Deployment


federicoweber:
Hey @msanroman snapshots tests are still failing, but I've decided to merge this in because the actuall build is working fine and it's solving all the style glitches we are seeing on production.

Will fix the tests in a new PR

PR #356, opened 2019-03-01
Move innerRefs to refs

msanroman:
### Purpose

With the move to `styled-components@4` we need to migrate `innerRef` to `ref`.


tiggreen:
After tests pass and the branch is working fine then let's ship it!

PR #359, opened 2019-03-08
Fix test suite

federicoweber:
### Purpose
To fix our test suite

**This is a temporary solution until we have time to properly upgrade to Babel 7.0 (that is required to properly work with styled components 3.4.10, cause of `React.forwardRef`).**

### Notes
What we are doing in here is copying `./lib` folder from `@bufferap/ui` `legacy-styled` branch, into our `shared-components` package. In this way, we can consume it by importing it with `@bufferapp/snalyze-shared-components/BufferUi`, and because it will behave exactly the same as importing the library from the package manager, it will be easy to switch once we move to Babel 7.0.

The reason I opted to have `BufferUi` library living in shared components is that I didn't want to create a short-lived package that is a copy of another one, and that felt like a good place to add it.

_Notice that we are not directly exporting `BufferUi` from `analyze-shared-components` to enforce the import with `@bufferapp/snalyze-shared-components/BufferUi`, and keep things nice and simple to replace._

#### Staging Deployment


msanroman:
Thank you, @federicoweber! 👏

PR #364, opened 2019-03-20
[WIP] Interface tweaks

moreofmorris:
### Purpose
I really wanted this pull request to be focused on just one area but I got excited and ended up touching quite a few things at once, sorry! The goal for this pull request is to update some of the visuals to freshen them up a bit. This doesn't include any swapping out of components from the Buffer Design System, yet!

Areas that this touches:
- Visual update to the sub tabs
- Colour change for Average Performance graph
- Colour change and removals from Metrics Breakdown graph
- Complete redesign of the Post Breakdown table
- Charts now have a little flexible width (up to 1008px) which is inline with the new BDS

http://task-interface-tweaks-analyze.dev.buffer.com/ :smile:_


moreofmorris:
Hey there @federicoweber, I wonder if you could take a look over this pull request when you had some time? I've just updated the snapshot tests and they did pass locally before pushing. For some reason the CI stuff seems to be taking forever and I need to go offline now. Could you take a look? Just a note, this doesn't include any swapping out of components from the new components library. This is a batch of visual updates using current styling only.

federicoweber:
Double checking on Instagram strange text spacing that hasn't been introduced in here but was standing out because of the new text trimming.
I think something like this `post.text.replace(/\n+/g,' ')` should help a bit.

Also @moreofmorris do you think we can add an extra line of visible text?

moreofmorris:
Thanks so much @federicoweber 👍 I've added in the changes (the line break/spaces fixer and an extra line so it's three) and I'll now merge 🤞

PR #368, opened 2019-03-29
New welcome box on Dashboard

moreofmorris:
### Purpose
We need an update to our welcome area on the Dashboard. This pull request is me trying to get one in place! This also moves Beamer down into the bottom left of the sidebar.

#### Staging Deployment
https://task-update-welcome-analyze.dev.buffer.com :smile:_


moreofmorris:
Hey there @msanroman, I wonder if you could take a quick look at this when you have a mo. Be great to get this out. It's a new welcome box and I moved Beamer down to the bottom left of the sidebar.

packages/web/components/DefaultPage/Welcome.jsx
```diff
@@ -0,0 +1,163 @@
...
+  
+  return greet;
+}
+
+function Welcome() {
+  return (
+    <Container>
+      <Inner>
+        <Contents>
+          <Title>
```
msanroman: What if we moved these nested components to its own file? The clearest example is Title, and how it composes the Title style, a certain Text style, and a greeting that's calculated with `new Date()`.

It might look like something like this:

```jsx
// File Welcome/Title.jsx
const Title = styled.div`
  margin: 0.25rem 0 0.375rem 1.5rem;
   span {
    font-size: 1.75rem !important;
  }
`;

export default () => {
  const now = new Date();
  const hours = now.getHours();
  let greet;

   if (hours < 12) {
      greet = 'Good morning';
  } else if (hours >= 12 && hours <= 17) {
      greet = 'Good afternoon';
  } else if (hours >= 17 && hours <= 24) {
      greet = 'Good evening';
  }

  return (<Title>
    <Text color="outerSpace" weight="semi-bold" size="extra-large">
      {greet}
    </Text>
  </Title>)
};
```

I think we could do the same with `ItemHeader` and `ItemDescription`, where we don't have to repeat the text styling on each instance:

```
// File Welcome/ItemHeader.jsx
const ItemHeader = styled.div`
  margin: 0 0 0.5rem;
`;

export default ({ children }) =>
  <ItemHeader>
    <Text color="outerSpace" weight="medium">
      {children}
    </Text>
  </ItemHeader> 
;
```

Then we can just use it as `<ItemHeader>Dashboard</ItemHeader>` in the Welcome box 😄 

How does that feel?

packages/web/components/DefaultPage/index.jsx
```diff
@@ -33,93 +21,30 @@ const Container = styled.div`
...
-
-    const reportsLink = (<Link
-      href="/reports"
-      unstyled
-      onClick={e => navigate(e, dispatch, '/reports')}
-    >your reports</Link>);
-
-    const welcomeText = (
-      <span>
-        Get started by viewing an <Link href="/overview" unstyled onClick={e => navigate(e, dispatch, '/overview')}>overview of your performance</Link> or gaining some insights around <Link href="/posts" unstyled onClick={e => navigate(e, dispatch, '/posts')}>your posts</Link>. You can also view {reportsLink}.
```
msanroman: I'm curious, we no longer want to add any navigation to this welcome text? 🤔

moreofmorris: In this situation it wouldn't make a ton of sense as this is more of an orientation piece now. As the descriptions are more about the concepts than actual destinations, if that makes sense. This welcome box is very much a v0.1 though, the next iteration will include "quick tips" which could well include links to actual destinations. Thanks for thinking about this though Mike!

packages/web/components/DefaultPage/Welcome.jsx
moreofmorris: Yep, sounds great, I'll do this now.

PR #370, opened 2019-03-29
remove https://k8s-api-php71.buffer.com reference

federicoweber:
### Purpose
Tp Remove https://k8s-api-php71.buffer.com reference
This isn's need anymore as we are fully migrated to it 😄.

#### Staging Deployment


tiggreen:
Awesome, thanks Fed!

PR #373, opened 2019-04-20
Add events tracking with Segment

federicoweber:
### Purpose
To track user events with Segment

### Notes
This adds the following tracking events:
- identify
- page export
- report created
- report module added
- report exported
- page viewed

#### Staging Deployment


federicoweber:
Hey @djfarrelly and @msanroman, this is a first pass at adding Segment tracking. As of now it's just identifying the user and tracking some of the events ( page change, create a report, add to report).

Planning to add the rest of the events by Monday, and also add the missing information in Identify.

packages/add-report/middleware.js
```diff
@@ -20,6 +39,14 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
       }));
       break;
     case actionTypes.CREATE_REPORT:
+      store.dispatch(analyticsActions.trackEvent('Report Created', {
+        product: 'analyze',
+        title: action.name,
+        userId: store.getState().appSidebar.user.id,
```
djfarrelly: @federicoweber, I think since the `identify` call will have happened with the user's "Account ID" we can remove adding the `userId` and the `organizationId` from all `trackEvent` calls.

packages/add-report/middleware.js
```diff
@@ -1,12 +1,31 @@
...
   switch (action.type) {
     case actionTypes.ADD_TO_REPORT:
+      store.dispatch(analyticsActions.trackEvent('Report Module Added', {
+        product: 'analyze',
+        userId: store.getState().appSidebar.user.id,
+        organizationId: store.getState().profiles.organizationId,
+        module: action.chart_id.replace('-', ' '),
+        reportId: action.reportId,
+        channel: isChartWithNoProfilesAttached(action.chart_id) ?
+          'none' :
```
djfarrelly: Can we just leave this to be `null` instead of a `none` string value? I think we should not pass this property if there is none

packages/add-report/middleware.js
```diff
@@ -1,12 +1,31 @@
...
+  const start = moment(date.startDate, 'MM/DD/YYYY');
+  const end = moment(date.endDate, 'MM/DD/YYYY');
+  return end.diff(start, 'days') + 1;
+}
+
 export default store => next => (action) => { // eslint-disable-line no-unused-vars
   switch (action.type) {
     case actionTypes.ADD_TO_REPORT:
+      store.dispatch(analyticsActions.trackEvent('Report Module Added', {
+        product: 'analyze',
```
djfarrelly: Do you think we should have an "analytics product constant" somewhere? Perhaps like 

```js
import { constants as analyticsConstants } from '@bufferapp/analytics-middleware';
// ...
product: analyticsConstants.PRODUCT_ANALYZE,
// ...
```
This could help save us from typos that might pollute our data in the future

packages/account/middleware.js
```diff
@@ -9,6 +12,11 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
           organizationId: action.result[0].organizationId,
         },
       }));
+
+      store.dispatch(analyticsActions.init(state.appSidebar.user.id, {
+        email: state.appSidebar.user.email,
```
djfarrelly: @federicoweber I've chatted with Eduardo about the `identify` and `Account Created` events and read a bit more in the Segment docs, I think we should skip passing the email here in Analyze.

packages/account/middleware.js
```diff
@@ -9,6 +12,11 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
           organizationId: action.result[0].organizationId,
         },
       }));
+
+      store.dispatch(analyticsActions.init(state.appSidebar.user.id, {
```
djfarrelly: This is the global account id or is the Publish user id?

packages/account/middleware.js
```diff
@@ -9,6 +12,11 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
...
+        email: state.appSidebar.user.email,
+        organizationId: action.result[0].organizationId,
```
djfarrelly: What Organization is this? The Publish org?

packages/add-report/middleware.js
federicoweber: Oh nice didn't realize that it makes sense.

packages/account/middleware.js
federicoweber: Yes this is the publish org

packages/account/middleware.js
federicoweber: This is the Publish user ID, didn't realized we also have a global one 😞.
Where can I get that from?

djfarrelly: Hey @federicoweber, with the Publish users merged into the Core Auth svc, we now have a Global account ID. You should be able to get this on the server side via the session service. Perhaps @esclapes could help with any questions. 

I think if Analyze's backend is asking for the global scope of the session ([see example](https://github.com/bufferapp/core-session-service#get)) you should be able to grab it there and sent it to the client, or perhaps in the RPC event for the publish user you could hit the Auth svc w/ the publish user id and get the global id.

packages/server/index.html
```diff
@@ -19,6 +19,12 @@
   </head>
   <body>
     <div id="root"></div>
+    <script>
+      !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t,e){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=e};analytics.SNIPPET_VERSION="4.1.0";
+        analytics.load("MV5uQCIyUw5MR9oqTGgzK9Goqj8SlDcP");
```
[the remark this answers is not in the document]
djfarrelly: @federicoweber, I've also created a dev key for local testing (`YDcr43WV9yiBxqc1dWAgTSROLBZRa2hX`). You can add it locally and test events using the debugger tool here: https://app.segment.com/buffer/sources/segment_dev_analyze/overview

After this is in prod you can check the debugger for this source: https://app.segment.com/buffer/sources/segment_analyze/overview

packages/analytics-middleware/middleware.js
```diff
@@ -0,0 +1,23 @@
...
+      if (window.analytics) {
+        window.analytics.identify(action.userId, action.payload);
+      }
+      break;
+    case actionTypes.TRACK_EVENT:
+      if (window.analytics) {
+        window.analytics.track(
+          action.eventName,
+          Object.assign({
+            product: window.PRODUCT_TRACKING_KEY,
```
federicoweber: I decided to push this in here, so we won't need to pass it around for all the tracked events.

federicoweber:
Hey @djfarrelly, will fix that test but it's unrelated to this changes.
Other than that I think we are ready to roll if this looks good to you.

packages/server/rpc/user/index.js
```diff
@@ -17,6 +17,6 @@ module.exports = method(
     })
       .then((result) => {
         const userData = JSON.parse(result);
-        return parseUser(userData);
+        return parseUser(userData, req.session);
```
djfarrelly: @federicoweber, I'm wondering if perhaps any access tokens might be in this session body? It might be better to just send the global account id here. What do you think?

packages/server/index.js
```diff
@@ -52,6 +52,8 @@ const beamerScript = `
 const isProduction = process.env.NODE_ENV === 'production';
 app.set('isProduction', isProduction);
 
+let segmentKey = 'YDcr43WV9yiBxqc1dWAgTSROLBZRa2hX';
```
djfarrelly: 👍 defaulting to the local key

packages/profile-selector/middleware.js
```diff
@@ -61,6 +62,20 @@ function getNewProfilePath({ router }, profile) {
   return `/${profile.service}/${tool}/${profile.id}`;
 }
 
+function getPageNameFromPath(path) {
```
djfarrelly: Nice - this takes `/twitter/overview/52bc7a190a0e32ec4c000022` and turns it into `"overview"` - it's pretty cool the codebase already has this!

packages/profile-selector/middleware.js
```diff
@@ -61,6 +62,20 @@ function getNewProfilePath({ router }, profile) {
...
+  const tool = getCurrentToolFromRoute(path);
+  if (tool) {
+    return tool;
+  }
+
+  const page = path.match(/^\/(.*)$/);
```
djfarrelly: In what case would this return a "page" - Reports or Aggregates?

packages/add-report/middleware.js
```diff
@@ -20,6 +30,11 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
       }));
       break;
     case actionTypes.CREATE_REPORT:
+      store.dispatch(analyticsActions.trackEvent('Report Created', {
```
djfarrelly: In the tracking plan we have `reportId` - I'm fine with omitting this for now and perhaps checking back in with Tom on if we need this right away. Is getting this report id hard to do here @federicoweber?

packages/server/rpc/user/index.js
federicoweber: Good one @djfarrelly, the reason I opted to pass in the entire session, is that this will keep all the knowledge about what we need to extrapolate in `parseUser`.

The `parseUser` is then taking care of exposing only the global account id https://github.com/bufferapp/buffer-analyze/blob/4dd896f728a2183b0c60576964dfefedf2317d36/packages/server/rpc/utils/parseUser.js#L4

packages/profile-selector/middleware.js
federicoweber: Yeah I added some of those in when we redesigned the URL structure 😄.

packages/profile-selector/middleware.js
federicoweber: that's right

packages/add-report/middleware.js
federicoweber: hey @djfarrelly, sorry should have mentioned this in the notes.
I didn't add that initially because at that point we didn't have an id yet.
But just realized that I moving the tracking after the creation confirmation step we can get that id back from the APIs. Just pushed the change.

packages/server/rpc/user/index.js
djfarrelly: Oh - sorry :) Good stuff with this!

PR #377, opened 2019-04-27
Add Video and Carousel Previews for IG posts

federicoweber:
### Purpose
To show videos in Instagram posts
image

### Notes
Hey @tomredman and @moreofmorris I noticed that the preview for IG video was broken (showing a white box), and digging a bit into it I figured out that there is an issue in the way we are saving the data, and we are effectively storing only the mp4 url.

This is just a first pass, theoretically, the video could be played but I've decided to hide the controls because in that small space it was a bit cramped.
image

My plan is to start fetching the thumbnail so that we can better fill the square format with a cover (the video is not that flexible in term of scaling), and to add in a custom Play / Pause interface.

Going to merge this in, in the meanwhile if it feels good to you.

#### Staging Deployment


moreofmorris:
Post Breakdown Icons

Hey there, @federicoweber. We discussed in the chat, but will chat about here too for transparency (and for Tom's benefit)... the Post Breakdown table should never try to do too much and the items inside there should really serve as a reference to the post. The user should be able to recognise what the post is and then see the metrics related to it, but no more really. As a separate story, I want to work on that post modal which a user can click to see the post in more detail, which they can then see the image, video or carousel in full as well as see more detailed metrics.

For this view though, let's keep things simple. We won't show any video controls, ideally we'd show the video thumbnail and that would fill up that whole square. Again, it's merely a reference to the post (in the near future, they'd click to open a modal to see the whole thing and play it back if needs be). We did experiment with that video placement so it'd be great to make that fill the square. Same with any carousel type images we have, too (we'd just show the first image).

So the user can understand what the post refers to, we can use some clear icons to help show what types the posts are. I made some simple icons here we could use. The movie one is the same as in the Story Breakdown so we could reuse that asset, it's just displayed differently here (I want to update for Stories, but that's a separate task). These are black icons that sit on a white circle. Here is the CSS of the circle of the icon:

```
width: 32px;
height: 32px;
background: rgba(255, 255, 255, 0.75);
box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.08);
```

Here is the Figma file: https://www.figma.com/file/HRIRJVBhl9rUIfghEhEWauJ0/Analyze-Post-Breakdown-Icons?node-id=1%3A2 and here is the Dropbox folder: https://www.dropbox.com/sh/ql1cc978bm8p9am/AABcqOsCoPqrD8nSrEpX3j6Na?dl=0

Hope that helps! 😊

federicoweber:
Thanks for brainstorming this with me and preparing the icons @moreofmorris.
We are ready to roll with it
image

PR #379, opened 2019-05-06
Hide hourly chart from Twitter Overview

tiggreen:
### Purpose
I'm thinking to remove the package entirely after the chart is removed from Publish. What's your opinion @federicoweber :) 
#### Staging Deployment


federicoweber:
Hey @tiggreen, yep once it's gone all around (including in reports) we can remove both the package and the rpc endpoint. Thanks for taking care of this 🙏

PR #389, opened 2019-05-30
use Analyze API for Instagram posts

federicoweber:
### Purpose
To allow new to Buffer users to see their Instagram posts.

#### Staging Deployment


packages/server/rpc/postsAnalyzeApi/index.js
```diff
@@ -0,0 +1,36 @@
+const { method } = require('@bufferapp/buffer-rpc');
+const rp = require('request-promise');
+
+function mapSortBy(sortBy) {
```
federicoweber: This is only needed until we migrate all post requests to the new API.

PR #391, opened 2019-06-03
New Quick Tips module for Home

moreofmorris:
### Purpose
Working on a new Quick Tips module for Home. This module allows for quick and easy tips that a customer can click through for fun and learning. There are some links that are a little more dynamic which allow to click through to sub tabs like Overview or Posts. It's a super MVP feature right now but we're hoping we can build on this in the future and make more sophisticated.

### Notes
The new module needed to know the social accounts of the user so it could direct them to things like Overview or Posts tab. I used the same code as Social Accounts Overview module. Is that ok?

#### Staging Deployment
https://task-quick-tips-analyze.dev.buffer.com :smile:_


packages/quick-tips/components/QuickTips/index.jsx
```diff
@@ -0,0 +1,251 @@
...
+
+    this.state = {
+      animation: false,
+      tip: 0,
+      tips: [
+        <span>
+          Making a Report is a great way to put together graphs and tables in any combination you like!
+        </span>,
+        <span>
+          Check out {getHomeLink()}. Its a quick overview of all your social accounts in one place!
```
federicoweber: Not a big one, how do you feel about refactoring this using template literal, to remove the extra `<span>`?
Something like `Check out ${getHomeLink()}. Its a quick overview of all your social accounts in one place!`

packages/quick-tips/index.js
```diff
@@ -0,0 +1,29 @@
+import { push } from 'connected-react-router';
+import { connect } from 'react-redux';
+import QuickTips from './components/QuickTips';
+
+function getComparableUsername(profile) {
+  return profile.username.replace('@', '').toUpperCase();
```
federicoweber: great catch 💯

packages/quick-tips/components/QuickTips/index.jsx
moreofmorris: Great, I'll do this now 👍

moreofmorris: Hey there, Feds. I try it out but was getting `[object Object]` in the browser, I think because the links that are getting constructed are actual React nodes and not just strings. I've decided to add some light formatting to these tips too so I've just reverted back to `<span>`, hope that's ok 👍

PR #394, opened 2019-06-04
Allow non-Publish signups to use Analyze independently

msanroman:
### Purpose

To allow non-Publish users to sign up and use Analyze independently :tada:

### To Do 
- [x] Get user account
- [x] Get profiles
- [ ] ~Get IG posts~ (not part of this PR)
- [ ] Display trial information


packages/server/rpc/profiles/index.js
```diff
@@ -14,8 +15,23 @@ function sortProfilesByUsername(profiles) {
...
+        const response = await authenticationService.getChannels({ accountId: req.session.global.accountId });
+        return response.map(profile => ({
+          id: profile._id,
+          avatarUrl: profile.avatar,
+          service: profile.serviceType,
+          timezone: profile.timezone,
+          username: profile.name,
+          organizationId: profile.organizationId,
+        }));
+      } catch (e) {
```
federicoweber: Do you think it could be worth to fail in here and let the error middleware kick in?

packages/server/rpc/profiles/index.js
```diff
@@ -14,8 +15,23 @@ function sortProfilesByUsername(profiles) {
 module.exports = method(
   'profiles',
   'fetch profiles',
-  ({ id }, req) =>
-    rp({
+  async ({ id }, req) => {
+    if (!!req.session.analyze.foreignKey) {
```
federicoweber: could be nice to abstract this for readability, maybe something like `isGlobalUser()`

packages/server/rpc/user/index.js
```diff
@@ -1,22 +1,37 @@
...
-        includes: 'avatar',
-      },
-    })
-      .then((result) => {
-        const userData = JSON.parse(result);
-        return parseUser(userData, req.session);
-      }),
+  async (_, req) => {
+    try {
+      if (!req.session.analyze.foreignKey) {
```
federicoweber: Same in here 😃

PR #400, opened 2019-06-07
Task/latest reports

dinostheo:
### Purpose

One of our goals for this cycle is to improve the first experience of our users. Upgrading and improving our Home page is part of that plan. [Here](https://paper.dropbox.com/doc/Analyze-Home-1.0-Specification--AeY9ZTnXR36i3RQIRrINea5pAg-NYp58qDzKjay1WvqhOLQA) is a paper that shows what we want “Analyze Home 1.0” look like. Adding “Latest Reports” component to the home page is one of the tasks we want to accomplish. 

### Notes

1. I am reusing the state of the list reports and therefore the sorting of the reports (in order to find the three latest) is done in the front-end using the `updated_at` property.
2. Reports link to the report page. This was not specified in the document.
3. Also the link text is `Report`. As I see it now, it might be misleading. This needs some input.
4. For empty table of reports I reused the `<NoData />` component. However, I passed different text as the `There are no analytics for this date range` did not seem relevant. It displays `There are no reports`. This needs input as well :)

#### Staging Deployment

https://task-latest-reports-analyze.dev.buffer.com/


packages/latest-reports/components/ReportsTable/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+  reports: PropTypes.array
+};
+
+const ReportsTable = ({ reports, loading, ...props }) => {
+  let content = null;
+
+  if (loading) {
+    content = <Loading active noBorder large />;
+  } else if (reports.length === 0) {
+    content = <NoData children={<Text>There are no reports</Text>}/>;
```
tiggreen: 👌🏼

dinostheo: Good one. I didn't even think about it :). I'll work on it.

dinostheo:
Pushed the changes for opening a report without page reload. There is also some stealth removal of unused code in the same component. I was expecting that those would have been exposed by eslint.

packages/latest-reports/components/ReportsTable/index.jsx
tiggreen: Oh, I was just pointing it out. I don't think there is anything here we need to change 😄

tiggreen:
Great thank you @dinostheo! I think James has some feedback on the design side of things too!

moreofmorris:
There are a few design/QA things I'd love to see @dinostheo 😊I wasn't sure where best to put them, I opted for the JIRA ticket as they're not directly related to the actual code you're submitting... _your code looks great!_ If we can get these design tweaks in then we should be good to go. I've added them to the JIRA ticket here: https://buffer.atlassian.net/jira/software/projects/AT/boards/69?selectedIssue=AT-168

dinostheo:
The changes that you can find in the 2 latest commits are based on @moreofmorris feedback in JIRA https://buffer.atlassian.net/jira/software/projects/AT/boards/69?selectedIssue=AT-168

More detailed:
- Rebased against latest master
- Renamed the component to recent-reports
- Adjusted component on text content, size and color (based on feedback)
- Did some stealth Readme documentation & package.json dependencies removal

NOTE: Git branch is still using the old naming `latest-reports`.

PR #404, opened 2019-06-10
Bump components to latest version and adjust font sizes across app

moreofmorris:
### Purpose
With trying to move the interface forward there is a great need for the `0.875rem` (14px) font size in our interface. We can't fully use the brand new BDS `<Text>` component yet as it doesn't support all the use cases we need. That's ok, as we're a more visual interface and there are bound to be sizes and weights that are not required by the other apps. Rather than wait for that, I wanted to bump our `buffer-components` (the old components package) up to the latest version as this includes the `0.875rem` size we need. It also takes us off the rather strange `.alpha` version we were on. Doing this means, at the very least, we're using the latest (older) `buffer-components` package that the other apps are using and we get access to the more standard font sizes.

### Notes
In an ideal world, we'd be using the very latest BDS `<Text>` component, but we're actually kinda quite way off from that. I need to go back to Tyler and see how we deal with Analyze's special text treatments and how that would work... but that's in the future. A quick bump and font size adjustment seems quick and still inline with overall goals at the mo 👍 

### Review
I'd love for this to be checked over to see if there are any super strange or weird font size issues. I've done quite a bit of frontend testing here and bumping to the latest seems to be working fine. I know its a lot of file changes but they should all be to do with font size readjustments.

#### Staging Deployment

https://task-update-buffer-components-analyze.dev.buffer.com :smile:_


packages/shared-components/GridItem/components/Value/index.jsx
```diff
@@ -4,7 +4,7 @@ import Text from '@bufferapp/components/Text';
 import TruncatedNumber from '../../../TruncatedNumber';
 
 const Value = ({ children, showPercentSign, smaller }) =>
-  <Text size={smaller ? 'large' : 'extra-large'} weight="bold" color="outerSpace">
+  <Text size={!smaller && 'large'} weight="bold" color="outerSpace">
```
dinostheo: If smaller is truthy this will result to `false`.

e.g. 
```
const smaller = true;
const result = !smaller && 'large';
console.log(result); // false
```

or 
```
const smaller = false;
const result = !smaller && 'large';
console.log(result); // 'large'
```

I don't know what is the intention here. I 'm just checking to see if this is correct that the possible 2 values here are either `false` or `'large'`

moreofmorris: In the `<Text />` component from the `buffer-components` package, you can pass in these sizes: `'extra-small', 'large', 'extra-large', 'mini', 'small'` but there isn't actually one for the regular/normal size (which would be `1rem` or `16px`). Due to having to change many font sizes across the app, this particular change actually would be for that regular/normal size which actually doesn't have a string name like the others. We could pass in a pseudo `regular` or `normal` for it to make more sense from the Analyze size, eg: `<Text size={smaller ? 'normal' : 'large'}`? But in reality the `<Text />` component in `buffer-components` wouldn't pick up that `normal` value. Happy to change to this something that makes more sense if this one doesn't 👍

dinostheo: That's good. I just wanted to validate that there was either an error or a conscious decision and apparently it's the second :)

PR #408, opened 2019-06-11
use appSidebar.user status for redirect check

federicoweber:
### Purpose
As we already have the user data in the `appSidebar` store, we can consume that instead of duplicating the state in the `profileLoader`.

### Notes
Hey @tomredman, I did a small refactor on this one, to keep the states a bit cleaner 😉.

#### Staging Deployment


tomredman:
Beautiful, thanks @federicoweber. Was looking for the `user` somewhere but missed this one!

PR #413, opened 2019-06-14
Stories breakdown tweaks

moreofmorris:
### Purpose
Exploring some small updates to make Stories Breakdown less busy looking. Want to bring it more inline with our new Recent Posts table. Also, want to fix bug with pagination and responsiveness. Also, includes updates to the usage of the `bufferapp/ui` components to fix some bugs and make selects work correctly.

### Notes
The stories currently don't resize to browser width and at the moment the pagination isn't working quite as it should. I ended up going with a slightly different approach where I separate out the stories (and the fillers) into five item page chunks. That allows me to use `20%` which works with its parent using CSS only. I changed how the images in the stories are rendered so they also resize properly. I tried to simplify the interface a bit and tried to refactor the code a bit whilst I worked on this. I also made a new shared component which is the little icon used for media types. I wanted to only include stories breakdown changes in this pull request so I'll apply that new shared component in a future pull request.

### Staging Deployment
https://task-stories-tweaks-analyze.dev.buffer.com


moreofmorris:
Hey there, @federicoweber, thanks for this. I'll merge this in now 👍

PR #424, opened 2019-06-24
New shared component for post media

moreofmorris:
### Purpose
This is to improve the "no media" situation for Posts Breakdown. I've created a new shared component which will deal with all post media and this will be used between Recent Posts and Posts Breakdown to avoid repeating code and having repeatable design between them. Now fully using my new MediaIcon component that I created last week so we no longer have inline SVG for these anywhere where there are posts or stories.

### Notes
I have impersonated a few of our customers to try this out to make sure it's working ok. I'd appreciate just digging around and making sure nothing is broken here between the Recent Posts module and the Post Breakdown module. Ideally, the Post Breakdown module should now show a blue box with a bit of the text inside rather than the rather horrid "No media" thing.

### Staging Deployment
https://task-improve-no-media-analyze.dev.buffer.com


moreofmorris:
Hey there @dinostheo, thanks again for the review and the comment. I've just added in some prop types, took some away and make the rendering a little more consistent for that new shared component. Great catch! Thanks Dinos! 🙌

PR #430, opened 2019-06-26
Send the average requests for Facebook to analytics-api

dinostheo:
### Purpose
Fetches the daily averages for Facebook from the analytics-api.
It is part of the following task.
https://buffer.atlassian.net/jira/software/projects/AT/boards/69?selectedIssue=AT-188


packages/server/rpc/average/index.test.js
```diff
@@ -129,18 +129,18 @@ describe('rpc/average', () => {
...
     expect(rp.mock.calls[1])
-      .toEqual([{
-        uri: `${process.env.API_ADDR}/1/profiles/${profileId}/analytics/totals.json`,
-        method: 'GET',
-        strictSSL: false,
-        qs: {
-          access_token: token,
-          start_date: expectedStart,
-          end_date: expectedEnd,
-          profile_id: null,
```
tiggreen: Hm, not sure why were we sending `profile_id` as `null` here?

dinostheo: Those tests they were mocking requests to the old API. Since I changed Facebook to point to the analytics API, they are not relevant anymore.
Usually I see the following code in the request:

```
profile_id: (shouldUseAnalyzeApi(profileService) ?
  profileId :
  null
),
```

which tells me that the `profile_id` prop is not required for the old API. Perhaps the token can be used to identify the profile?

PR #431, opened 2019-06-26
Use CSS grid for layout

moreofmorris:
### Purpose
There were layout issues in Firefox and rather than trying to patch up a rather broken mess of styles and confusing semantics I wanted to go back and fix it more at the root here. Want to use a clearer way to layout the interface using CSS Grids which should hopefully fix the Firefox problem but also provide a much more readable codebase for making adjustments in the future. There are quite a few changes in this one and a lot of them affect the way the pages and tabs get rendered, this is the full run down of what this pull request does:
- fixed Firefox issues
- implemented Grid CSS
- created a new set of shared components to help with layout
- made interface work all the way down to smaller 1280px screen resolutions (properly!)
- extended maximum size of content to work on standard screens of 1440px screen resolutions

### Notes
Sorry again for all the file changes! There is a lot of refactoring here too and some semantic word changes to make everything easier to work with. I hope that working with the layouts in Analyze will be much, much easier moving forward.

### Staging Deployment
https://task-use-css-grid-for-layout-analyze.dev.buffer.com


dinostheo:
Looks good @moreofmorris 🚢

PR #433, opened 2019-06-27
Task/appcues integration

ivanaszuber:
### Purpose
To install the AppCues service in order to be able to build Product Tours within AppCues

### Notes

- Adds the appcues script to index.html
- changes the App component to be a class component and adds `componentDidUpdate` to call the `AppCues.page()` function on every route change
- adds `AppCues.identify()` calls and passes basic user information

The integration can be tested by opening https://task-appcues-integration-analyze.dev.buffer.com/?hey_appcues

This will trigger the AppCuess Dev tools which should show something like this:
Image 2019-06-10 at 10 44 41 AM

#### Staging Deployment
https://task-appcues-integration-analyze.dev.buffer.com :smile:_


ivanaszuber:
Yeah, great one @msanroman ! I do hope we actually move to the Segment integration though, to install AppCues via Segment, we can then remove this code 😅 . We've been having some issues with that, and since Alfred needs this by the end of the week to get started with the Product Tours, we decided to take this route and then switch to the Segment approach in a few weeks.

I'll make sure to clean up (remove) this code when we do 🙌

PR #443, opened 2019-07-09
Send the selected profile service on posts set_date_range

dinostheo:
### Purpose
Twitter requests for posts, when you set a date range will end up to analytics api, which would respond with a 400 HTTP code. 
The reason was that we would not pass the service on that action. Facebook and Instagram were not affected as analytics api extracts the service from the profile data. However, for twitter we want to fetch the posts from the buffer-web api and there we needed the service value in order to do the distinction in the express server.

This is reported here https://buffer.atlassian.net/browse/ADVO-6193

#### Staging Deployment

https://task-fix-twitter-posts-date-selection-analyze.dev.buffer.com/


dinostheo:
I will skip review for this as it's a very minor change.
/skip-review

PR #445, opened 2019-07-09
Show last time a profile has been updated [AT-183]

federicoweber:
### Purpose
To display the last time the data for a given profile has been updated.

#### Staging Deployment


moreofmorris:
Thanks so much for your speedy work on this @federicoweber! 🙌 I wonder if these little tweaks wouldn't be too much bother? Could we lower case the rest of the text so it reads more like `Last updated 4 hours ago` rather than it being capitalised and I wonder if we could slip a little time icon in before the text too. See this Figma design file here I quickly mocked up here: https://www.figma.com/file/TpuH1vY4jmhtCXF1g7AzM0XT/Analyze-Last-Updated?node-id=3%3A7

That time icon should be available to use already as its used in the Recent Posts module, for example.

PR #448, opened 2019-07-10
Use "profileService" instead of "service" for the /posts RPC request

msanroman:
### Purpose

All the RPC endpoints are using `profileService`, whereas the `/posts` one was using `service`. This fixes it!


federicoweber:
Great catch 💯

PR #452, opened 2019-07-12
Task/billing details modal

dinostheo:
### Purpose
This PR is quite big! However, there are a lot of components that are brought to this package from the buffer-account project. I will describe those under the NOTES section.

This PR implements a Billing details modal, which is triggered when the users click on the button on the trial banner. They can fill in their credit card information and the banner will be hidden.
For further subscription actions they have to go to the account page 

https://buffer.atlassian.net/jira/software/projects/AT/boards/69?selectedIssue=AT-216 

### Notes
By copying the billing-details component, I needed to bring along some ui components. Those components were not in the `@bufferapp/ui` library. 
Additionally, the billing-banner becomes part of the billing-details component and replaces the plan component from the buffer-account billing-details.
Finally, the logic that is related with canceling and renewing a subscription are excluded as they are not needed for the analyze client.

I would like to write more tests, but I thought to push the PR as it's quite big already.

### Review

There are stories available on storybook for seeing the component visuallly.
You can also try it on staging as well. 
Staging is using the live mode tokens of Stripe, so the test credit card won't work. But it's functional with a real credit card 💰 

#### Staging Deployment


tomredman:
Hey @dinostheo this flow feels great!

**1) Let's change the banner text.**
From this:
`You have 11 days remaining on your Analyze trial. You'll be billed $50 per month at the end of your trial.`

To this:
`You have 11 days remaining on your Analyze trial. Add your billing details to enjoy access after your trial.`
Screen Shot 2019-07-12 at 1 15 07 pm

**2) Minor modal updates**
This is how I'm seeing the modal now:
Screen Shot 2019-07-12 at 1 17 03 pm

Can we adjust it so it looks like this:
Screen Shot 2019-07-12 at 1 16 19 pm

Things to note:
- the header height should remain static (current grows/shrinks with window size): http://hi.buffer.com/31b8a47ea4c5
- "$50 billing at the end of your trial": `color:#598016`, `font-weight:500`

PR #455, opened 2019-07-15
Add buffer-product package for Cross-selling in the AppShell

CoxyBoris:
### Purpose
Heya Analyze team 👋 , I discussed this with @federicoweber on Friday already.
In order to do some cross-selling with the AppShell component, all products frontend need to have access to the global account from Core (which is only available from backends within the K8s cluster).

I created a small [npm package](https://github.com/bufferapp/respond-tools/tree/master/buffer-products) to import on all products backend. This is temporary until Core provides a public api, but needed now by the AppShell component.

I'm not really sure how to test this PR so feel free to change /edit anything and your help will be very appreciated! 😅

### Notes
I'll cc @ssvmvss who will be working on updating the appShell component

#### Staging Deployment


federicoweber:
hey @CoxyBoris, thanks for preparing the PR 🙏 .

One thing, I noticed it's failing to get the package from npm https://travis-ci.org/bufferapp/buffer-analyze/builds/559016929#L467-L470, any idea why?

CoxyBoris:
Great one @federicoweber, I think it fails because the build script needs the @respond npm token to be able to pull from npm 😅 I'm not sure how the build works for analyse, I'll reach out on Slack #engineering in the meantime to see if there is a quick fix for that. Otherwise I can always publish the package to the @bufferapp npm account.

federicoweber:
Hey @CoxyBoris, got it. As we are going to use it across all products my hunch is that we could move this under the **@bufferapp** namespace. What do you think?

CoxyBoris:
Hey @federicoweber , since nothing is sensitive we decided that this package could be public 😀 so it's solving the issue hehe. Looks like all the checks have passed? (probot/minimum-review-bot is still pending at the time I write this but not sure what it does 😅)

PR #458, opened 2019-07-19
Restructure to support banners in interface

moreofmorris:
### Purpose
This is an attempt to fix the layout scrolling issues where there is a banner present.

### Notes
The Analyze interface is laid out a bit different to our other products and relies heavily on CSS Grid and Flexbox. When adding other structural elements to the interface we have to consider the slightly more complex CSS that's in place. Pretty easy fix though, I'm hoping!

### Staging Deployment
https://task-fix-banner-scrolling-analyze.dev.buffer.com/


moreofmorris:
I don't really understand why the code coverage stuff is failing for me. I've taken away code in there so I would have thought test coverage would have increased? I don't really understand that bit of it 😅

PR #463, opened 2019-07-24
fix date range limit

federicoweber:
### Purpose
To fix the selectable date range for Facebook and Instagram profiles.

#### Staging Deployment


dinostheo:
Thanks for fixing that @federicoweber 🙏 . It was my mistake. I focused on fixing the requests bug and didn't noticed the breaking UI. Sorry, for that 😞 

Questions for @federicoweber & @msanroman 
I'm wondering though if returning null is fine for Facebook and Instagram, what would be the impact to do the same for Twitter? And that would be fine, is it an idea to dispatch an non fetch action from the date-picker middleware to instantiate the date picker and delete this rpc endpoint?

msanroman:
Good one. Would be a good one to chat with Tom about getting rid of this completely.

I'd say we could allow users to select any date range they want, and find out then if they have data for those dates or not, and that this just adds complexity right now 🤔Also, if we don't need to do an RPC request anymore to get the initial state of the calendar, it will just mean we can instantiate it through the default state of the reducer 😄

PR #466, opened 2019-07-25
adds a quick tip from Rothy's Social Media Lead

ivanaszuber:
### Purpose
Adds a new quick tip from the Rothy's social media lead, as per the marketing request here: https://threads.com/34368861958

#### Staging Deployment

https://task-add-new-quick-tip-analyze.dev.buffer.com :smile:_


msanroman:
Wow, so cool! :shipit:

PR #472, opened 2019-07-26
verify access token

federicoweber:
### Purpose
To verify access token and redirect to login service if it's wrong

#### Staging Deployment


packages/server/middlewares/verifyAccessToken.js
```diff
@@ -0,0 +1,41 @@
...
+    .catch(r => false);
+}
+
+module.exports = async (req, res, next) => {
+  // As this is a quite expensive check to perform, and we want the redirect to happen
+  // before the app is rendered, we are processing it only on the first request.
+  if (isRequestingApp(req)) {
+    const isValid = await isAccessTokenValid(req);
+    if (!isValid) {
+      return redirect(res, 'https://login.buffer.com/login?redirect=https://analyze.buffer.com');
```
msanroman: Does this mean then that `convert_access_token` has removed the session/access token if invalid? We don't need to log them out by sending them to `/logout`? :)

federicoweber: we won't need logout, visiting the login service will reset the token :)

federicoweber:
Hey @philippemiguet I think the middleware in here is working as expected but there is something in `login` that is manipulating the redirect for Analyze. I tested that changing the redirect parameter to `publish.buffer.com` and even `foo.buffer.com` and the redirect is working as expected, if I set it to `analyze.buffer.com` I'm getting redirect to `https://account.buffer.com/analyze`

philippemiguet:
Hey @federicoweber great to hear, I will check the login service to understand what is going on. @msanroman @federicoweber, do you have an email/password for an old Analyze users (using Publish) which I can use to test?

philippemiguet:
@federicoweber I wonder if it works, re-issue tokens then hit verifyAccessRights which then redirect to Analyze. An easy way to verify that is to open the network tab in Chrome, check preserve log and run the test. Also, we could add console.log in both middleware to understand.

I've checked the login service, and I don't find how we could send you to the Account app with the above workflow.

PR #483, opened 2019-08-01
Task/stories summary

dinostheo:
### Purpose

Implements Instagram stories summary table as per https://buffer.atlassian.net/jira/software/projects/AT/boards/69?selectedIssue=AT-52
It is based on the following design document
https://paper.dropbox.com/doc/Instagram-Stories-1.0-Specification--Ah8iQPXXbs8dvEghtY39aGENAg-KjeMCgJtQJv566Sx8t77C

### Notes
It depends on the https://github.com/bufferapp/buffer-analytics/pull/410 in order to get the data.

### Review
You need to point to the staging analytics-api from the PR above in order to get the data and be able to test on staging.

#### Staging Deployment

https://task-stories-summary-analyze.dev.buffer.com/


packages/server/rpc/storiesSummary/index.js
```diff
@@ -0,0 +1,77 @@
...
+module.exports = method(
+  'stories_summary',
+  'fetch instagram stories summary',
+  ({ profileId, startDate, endDate }, req) => {
+    const dateRange = new DateRange(startDate, endDate);
+    const previousDateRange = dateRange.getPreviousDateRange();
+    const analyzeApiAddr = req.app.get('analyzeApiAddr');
+    const accessToken = req.session.analyze.accessToken;
+
+    const currentPeriod = requestStoriesTotals(accessToken, dateRange, profileId, analyzeApiAddr);
```
federicoweber: Hey @dinostheo, given that we are adding a new API endpoint just for this, and that we won't be consuming stories totals without showing the difference with the previous period, do you think we could directly compute the difference in the APIs?

That could save us a request, and simplify things a bit by maintaining all the data manipulation in one place.

packages/server/rpc/storiesSummary/index.js
```diff
@@ -0,0 +1,77 @@
+const { method } = require('@bufferapp/buffer-rpc');
+const rp = require('request-promise');
+const DateRange = require('../utils/DateRange');
+
+const LABELS = {
```
federicoweber: Not a big one, but do you think we could do this on the chart itself?

packages/server/rpc/storiesSummary/index.js
```diff
@@ -0,0 +1,77 @@
...
+  engagementRate: 'Engagement Rate',
+};
+
+const requestStoriesTotals = (
+  accessToken,
+  dateRange,
+  profileId,
+  api
+) => rp({
+    uri: `${api}/metrics/stories_totals`,
```
federicoweber: I think we can directly set the address in here `${req.app.get('analyzeApiAddr')}/metrics/stories_totals`, instead of passing it as an argument 😉.

packages/server/rpc/storiesSummary/index.js
dinostheo: @federicoweber That's a very good point 👍. Actually do it this it will automatically deprecate the following comments as well.

PR #484, opened 2019-08-01
fix profiles rpc

federicoweber:
### Purpose
This is using `user._id` instead of `user.id` to get profiles for a given user

#### Staging Deployment


tomredman:
Thanks for this fix @federicoweber! Did something change that required the ‘_id’ instead of ‘id’? Will this continue to work for both new users and legacy users?

federicoweber:
hey @tomredman, this was needed cause we were using `_id` for new users and `id` for legacy users, see https://github.com/bufferapp/buffer-analyze/pull/480#pullrequestreview-269075874

PR #485, opened 2019-08-06
Better data refresh strategy [WIP]

federicoweber:
### Purpose
To refresh all relevant data without reloading the page.

#### TODO
[] trigger refresh from all middlewares that do not require a single profile
[] trigger refresh from all charts middleware if the current profile has fresh data

### Notes
This also has a better notification message that details which social account has fresh data

image

#### Staging Deployment


federicoweber:
Hye @tomredman how does the copy looks to you?

tomredman:
Looks great @federicoweber, thanks for going ahead and shipping this! The only I spotted in the screenshot was the extra space after the command "," shouldn't be there, but it looks like you might have fixed that in the code already.

PR #490, opened 2019-08-29
[AT-261] Replace inputs with BDS inputs

gmzjuliana:
### Purpose
Change Report Title and description inputs for BDS inputs

### Notes
I had a bit of trouble adding the tests for the handleBlur function in editDescription but I was thinking it might not be worth investing the time because once I've got the textArea component ready in BDS we'll be switching that out anyways, happy to dig in more if it makes sense to get them working though!

#### Staging Deployment
https://task-change-inputs-bds-analyze.dev.buffer.com/


packages/report/components/EditDescription/index.jsx
```diff
@@ -45,14 +57,24 @@ class EditDescription extends Component {
...
+        <InputWrapper>
+          <Input
+            forwardRef={this.textInput}
+            value={this.state.description}
+            onChange={this.handleChange}
+            onBlur={this.handleBlur}
+            autoFocus
+          />
+        </InputWrapper>
+        <ButtonWrapper ref={this.clearButton} >
```
federicoweber: We already chatted about this but leaving a not in here for transparency 😄.
I think this works fine for now, but eventually, we want to refactor the `Button` component in BDS to allow reference propagation.

gmzjuliana:
> Hey @gmzjuliana great work on this one, definitely a tricky one with the ref issue for the clear 🌮 .
> 
> I think it's ok to skip the test for the `handleBlur`, but feel free to reach out if you want to brainstorm on that one.
> 
> Only thing I noticed is that pressing **enter** while editing the description field is causing the page to reload without saving the description; that is working as expected on the title field 😄.
> 
> Feel free to merge this one once that issue has been addressed 🙌 .

Added an onSubmit to the form to prevent the reload :)

PR #491, opened 2019-09-03
Architecture review: In-app permissions

msanroman:
Proposed here is the architecture for handling permissions throughout our app.

While working on https://github.com/bufferapp/buffer-analyze/tree/task/add-analyze-pro-plan I've been exploring different options as I was experiencing some initial coupling between the logic for rendering different features and our plans, and I set this to share early thoughts about it.

**Note:** This is a first stab at doing architecture reviews asynchronously and leaving the final decision in our repo. Changes on the architecture could go through the pull review flow as any other piece of code, and we could have a centralised source of truth. Let me know how this feels!

---

# In-app permissions
This file aims to lay down the initial design and architecture on how to handle in-app permissions within this codebase.

## Goal
We want to prevent the user from performing certain actions, or rendering some charts, depending on their current subscription plan.

### Rendering components based on permissions
In the past, we’ve dealt with permissions with code that looked like the following pattern:

```javascript
// Rendering a component
render() {
  return (
     {this.props.userIsFoo && (<ProOnlyComponent />)}
  )
}
```

What happens if there’s dozens of these instances throughout the codebase, and then we make the decision to change which plans (and under which conditions) these accesses still apply?

We are also coupling each component with knowledge about the plan, so anytime we change a component, we might affect the plan-based features, and viceversa: a change in plan management might affect several components.

Let’s say I I just want to display Stories analytics only for Premium accounts. For simplicity sake, let’s say I have a component named `<StoriesAnalytics>` that renders everything.

I could take the legacy approach and say:

```javascript
// Rendering a component
render() {
  return (
     {this.props.isPro && (<StoriesAnalytics />)}
     {!this.props.isPro && (<UpgradePrompt />)}
  )
}
```

But this introduces the coupling I was mentioning earlier: my code in charge of rendering stories analytics now knows a ton more about how to retrieve the user’s pricing plan too!

I propose decoupling it by changing the approach we are taking.  What if we would just allow/disallow a certain component for a given user by checking if they have access to a feature?

```javascript
// Rendering a component
render() {
  return (
     <Can
        action={SEE_STORIES_ANALTICS}
        yes={<StoriesAnalytics />}
        no={<UpgradePrompt />}
      />
  )
}
```

`Can` here is a component that checks the given permission against the current set of permissions enabled for the given account in the application state. Then it makes the decision to render the `yes` component if the account has permission for it, or the `no` component otherwise.

`Can` can be a redux connected component like:

```javascript
export const Can = connect(
  state => ({
    permissions: state.permissions,
  }),
)(({ action, permissions, yes, no }) => (
  permissions.includes(action) ? yes : no
);
```

### Storing the permissions

How does the `Can` component get these permissions? We can have our account reducer handle that! When we receive our billing information in the frontend, we can decide which set of permissions we want to set:

```javascript

export const STORIES = 'stories';
export const LOGO_UPLOAD = 'logo upload';

export const PRO_PLAN = 'pro';
export const PREMIUM_PLAN = 'premium';

export const PERMISSIONS_BY_PLAN = {
  [PRO_PLAN]: [],
  [PREMIUM_PLAN]: [STORIES, LOGO_UPLOAD],
};

const initialState = {
  permissions: [],
}

export default (state = initialState, action) => {
  switch (action.type) {
    case BILLING_DETAILS:
	    return {
        ...state,
        permissions: PERMISSIONS_BY_PLAN[action.plan]
      }
    default:
      break;
  }
}
```

This would ensure we have the code that handles the permissions based on plan (or if in the future we were to introduce more complex, or dynamic, permissions) in a single point, while still being decoupled from the actual components and the UI logic that is enabled or disabled by those permissions.


federicoweber:
Hey @msanroman, thanks for starting this.

Overall I really like the proposed implementation, having a centralized permission handling feels like the best option, and having the permission stored on the redux level allows for some great interplay with all levels of the flow. For example, if we have the needs for it, we can easily add in a middleware to intercept actions and RPC requests.

My only observation is with the structure of the React Component itself:
I'm thinking we could split this into two components(`<Can />` and `<Cant />`); passing the conditions directly as props, and the content to render as a children.

This approach will make it really easy to read what is going on:
`<Can SEE_STORIES_ANALTICS><Stories /></Can>`
`<Cant SEE_STORIES_ANALTICS><UpgradeMessage/></Cant>`.

And we could also check for multiple permissions at once, that could be useful for example to only display an upgrade message if a user can't see any of the components of the page.

```
<Page>
    <Can SEE_STORIES_ANALTICS><Stories /></Can>
    <Can SEE_POSTS_ANALTICS><Stories /></Can>
    <Can SEE_SUMMARY_ANALTICS><Stories /></Can>
    <Cant SEE_STORIES_ANALTICS SEE_SUMMARY_ANALTICS SEE_POSTS_ANALTICS><UpgradeMessage/></Cant>
</Page>
```

The component itself would be something like this:
```
class Can extends React.Component {
  validatePermissions(props) {
    return Object.keys(props)
      .filter(key => key !== 'children')
      .every(permission => VALID_PERMISSIONS.includes(permission));
  }
  render() {
    return (<div> {this.validatePermissions(this.props) && this.props.children} </div>);
  }
}
```

dinostheo:
@msanroman This is great 💯 
This type of documents can help as well in the onboarding of new members or for ourselves a few months further when we try to figure out why we have taken certain decisions. 

I have only one question regarding permissions.
Are those also stored and served from the server side? e.g. Would a user that has no permissions to certain charts be able to get the data from the API and then the client will only hide this functionality, or is it additional to a server side check?

msanroman:
> Hey @msanroman, thanks for starting this.
> 
> Overall I really like the proposed implementation, having a centralized permission handling feels like the best option, and having the permission stored on the redux level allows for some great interplay with all levels of the flow. For example, if we have the needs for it, we can easily add in a middleware to intercept actions and RPC requests.
> 
> My only observation is with the structure of the React Component itself:
> I'm thinking we could split this into two components(`<Can/>` and `<Cant/>`); passing the conditions directly as props, and the content to render as a children.
> 
> This approach will make it really easy to read what is going on:
> 
> ```js-jsx
> <Can SEE_STORIES_ANALTICS><Stories/></Can>
> ```
> 
> ```js-jsx
> <Cant SEE_STORIES_ANALTICS><UpgradeMessage/></Cant>
> ```
> 
> And we could also check for multiple permissions at once, that could be useful for example to only display an upgrade message if a user can't see any of the components of the page.
> 
> ```js-jsx
> <Page>
>     <Can SEE_STORIES_ANALTICS><Stories/></Can>
>     <Can SEE_POSTS_ANALTICS><Posts/></Can>
>     <Can SEE_SUMMARY_ANALTICS><Summary/></Can>
>     <Cant SEE_STORIES_ANALTICS SEE_SUMMARY_ANALTICS SEE_POSTS_ANALTICS><UpgradeMessage/></Cant>
> </Page>
> ```
> 
> The component itself would be something like this:
> 
> ```js-jsx
> class Can extends React.Component {
>   validatePermissions(props) {
>     return Object.keys(props)
>       .filter(key => key !== 'children')
>       .every(permission => VALID_PERMISSIONS.includes(permission));
>   }
>   render() {
>     return (<div> {this.validatePermissions(this.props) && this.props.children} </div>);
>   }
> }
> ```

Good one @federicoweber! I've given this a try on [this commit](https://github.com/bufferapp/buffer-analyze/commit/68fc1b6f3e05860536e5b026fd4849826a82f65a) and seems to work great! Will update the docs here 👍

msanroman:
> Are those also stored and served from the server side? e.g. Would a user that has no permissions to certain charts be able to get the data from the API and then the client will only hide this functionality, or is it additional to a server side check?

Great question @dinostheo! Not for now. Even in the frontend, we are just "hiding" the data, but we are still requesting it and displaying it behind an opacity layer. In the future we could be more restrictive, but for now we are somehow "teasing" this experience with their real data anyway :)

msanroman:
Tweaked the docs with `Can` and `Cant` [here](https://github.com/bufferapp/buffer-analyze/pull/491/commits/7f828a9ad4a324c4795d196d4acdabe253d59508). If it all looks good to  you by the time of our weekly sync tomorrow I'll merge this one! :)

PR #496, opened 2019-09-08
[CORE-769] Update billing details package to use setup intent and sca

oscoda:
### Purpose

Adds Stripe Elements workflow with SCA.

#### JIRA
[CORE-769](https://buffer.atlassian.net/browse/CORE-769)

### Review

Screen Recording 2019-09-08 at 10 20 AM

#### Staging Deployment


packages/billing-details/reducer.js
```diff
@@ -6,6 +6,11 @@ export const actionTypes = keyWrapper('BILLING', {
   OPEN_MODAL: 'OPEN_MODAL',
   CLOSE_MODAL: 'CLOSE_MODAL',
   SAVE_CARD: 'SAVE_CARD',
+  CREATE_SETUP_INTENT_REQUEST: 'CREATE_SETUP_INTENT_REQUEST',
+  CREATE_SETUP_INTENT_SUCCESS: 'CREATE_SETUP_INTENT_SUCCESS',
+  HANDLE_CARD_SETUP_REQUEST: 'HANDLE_CARD_SETUP_REQUEST',
```
oscoda: @philippemiguet Just one quick note here that I noticed, but these that start with `HANDLE_CARD_SETUP_` should be `HANDLE_SETUP_CARD_` to match with that's in the actions below 🤦 (okay I'm leaving now for the week 😁)

packages/billing-details/reducer.js
```diff
@@ -16,6 +21,10 @@ const initialState = {
   reenabling: false,
   plan: null,
   trial: null,
+  creatingSetupIntent: false,
+  setupIntentClientSecret: '',
+  validatingCardSetup: false,
+  handleCardSetupError: '',
```
philippemiguet: @oscoda, I think there is no need to create a new state and error field for each action. Here we can re-use `error` and `validating`. It makes things easier down the line :)

packages/billing-details/reducer.js
```diff
@@ -40,6 +49,33 @@ export default (state = initialState, action) => {
         validating: false,
         modalOpen: false,
       };
+    case actionTypes.CREATE_SETUP_INTENT_SUCCESS:
+      return {
+        ...state,
+        setupIntentClientSecret: action.setupIntentClientSecret,
+        creatingSetupIntent: false,
```
philippemiguet: @oscoda Same as above for all this section, let's re-use the `error` and `validating` state :)

packages/billing-details/middleware.js
```diff
@@ -11,19 +11,29 @@ export const ERROR_CODES = {
...
-              token: response.id,
-            },
-          }));
-        }
-      });
+      store.dispatch(
+        asyncDataFetchActions.fetch({
+          name: 'updateCardInformation',
+          args: {
+            token: action.card,
```
philippemiguet: Since we renamed the param in the RPC endpoint, we should pass `{ paymentMethodId : action.paymentMethodId}` here :)

packages/billing-details/components/CreditCardForm/form.jsx
```diff
@@ -0,0 +1,49 @@
...
+import { CardElement, injectStripe } from "react-stripe-elements";
+import { Button } from "@bufferapp/ui";
+import { ButtonWrapper } from "./styles";
+
+class CreditCardForm extends Component {
+  handleSubmit = ev => {
+    ev.preventDefault();
+
+    const { setupIntentClientSecret, stripe, closeAction, save } = this.props;
+    stripe
```
philippemiguet: @oscoda - we should move this to the middleware to have this component as dumb as possible. This doesn't need to be executed in the form. As long as we use the `stripe` instance from the props, which we can pass to the middleware we are fine :)

packages/server/rpc/billing/updateCardInformation.js
```diff
@@ -33,18 +33,18 @@ function addCardDetails({ organizationId, product, stripeCustomerId, token, acco
...
+  async ({ paymentMethodId }, req) => {
     const adminAccountId = req.session.global.accountId;
     const { _id, stripeCustomerId } = await getOrganization({ adminAccountId });
 
     await addCardDetails({
       product: 'ANALYZE',
       organizationId: _id,
       stripeCustomerId,
-      token,
+      paymentMethodId,
```
philippemiguet: @oscoda - we probably need to update the addCardDetails function to make sure we send the right param to the Publish API :)

PR #503, opened 2019-09-17
Task/chart notes

gmzjuliana:
### Purpose
Add a way for users to include chart notes to their charts
https://paper.dropbox.com/doc/Design-Brief-Spec-Analyze-Report-Chart-Notes-bbBrIvGNRkBs5FFL6FpI1
https://buffer.atlassian.net/jira/software/projects/AT/boards/69?selectedIssue=AT-261

#### Staging Deployment
https://task-chart-notes-analyze.dev.buffer.com/


packages/report/middleware.js
```diff
@@ -131,6 +132,20 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
         },
       }));
       break;
+    case actionTypes.SAVE_CHART_NOTE:
```
federicoweber: Hey @gmzjuliana the way this is designed we are sending a lot's of events ( every time we open or blur note fields) even if no edit is performed.

I think we want to limit some of those and limit fetch and tracking to the cases where the note is actually different from the one stored. We can do comparing the note passed in the action with the one present in the state.

Let me know if you want to chat about this :).

packages/server/rpc/index.js
```diff
@@ -27,7 +28,7 @@ const profilesOverview = require('./profilesOverview');
 const getAccountDetails = require('./getAccountDetails');
 const hashtags = require('./hashtags');
 const aggregate = require('./aggregate');
-const lastUpdated = require('./lastUpdated');
+// const lastUpdated = require('./lastUpdated');
```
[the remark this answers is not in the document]
gmzjuliana: oh no! sorry about this 🙈 this was a temp fix that Tigran and I implemented for the hot reloader issue in my local

packages/report/reducer.js
```diff
@@ -58,6 +60,32 @@ const getLogoUrl = (report) => {
   return report.logo.url ? report.logo.url : '';
 };
 
+// abstract this into 1 getchart.. something that consolidates this in 1
```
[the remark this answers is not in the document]
gmzjuliana: 🤯wowwww, so cool!!

packages/report/middleware.js
gmzjuliana: heya @federicoweber I added in a check to see if it's the same note and if it is, it breaks.. how does this feel now?

packages/report/middleware.js
```diff
@@ -20,11 +21,14 @@ const dateRangesDoNotMatch = (reportDate, dateRange) => (
   (!dateRange.preset && reportDate.range)
 );
 
+const currentChart = (charts, chartId) => charts.find(item => item._id === chartId);
```
federicoweber: Two small ones :).

I prefer to use the function declaration instead of assigning the method to a class, as to me it seems a bit cleaner https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function.

For naming a good rule of thumb is to try to explain what a method does with is name, in this case this is  filtering charts, and return the one that matches the id so maybe something like `getChartById`.

packages/report/middleware.js
federicoweber: this is great thanks 🙏

packages/report/reducer.js
```diff
@@ -58,6 +60,15 @@ const getLogoUrl = (report) => {
...
```
federicoweber: I think we can clean up this comment 😄

packages/report/reducer.js
```diff
@@ -58,6 +60,15 @@ const getLogoUrl = (report) => {
...
+const editChartById = (charts, chartId, data) => charts.map((item) => {
```
federicoweber: so much cleaner 😍

packages/report/reducer.js
gmzjuliana: oops! yes! ahaha

packages/report/middleware.js
```diff
@@ -131,6 +137,25 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
         },
       }));
       break;
+    case actionTypes.SAVE_CHART_NOTE:
+      // if the note hasn't changed from what we have, break to avoid multiple calls
+      chart = getChartById(state.report.charts, action.chartId);
+      if (action.note === chart.note) {
+        break;
+      }
+      store.dispatch(analyticsActions.trackEvent('Report Chart Note Edited', {
```
federicoweber: hey @gmzjuliana I just noticed we need to add another property for the tracking data:
```
store.dispatch(analyticsActions.trackEvent('Report Chart Note Edited', {
    product: 'analyze',
    module: action.chartId.replace('-', ' '),
    reportId: state.report.id,
}));
```

gmzjuliana: gotcha! will do :)

PR #509, opened 2019-09-26
Deprecate analytics_start_date rpc

dinostheo:
### Purpose
We are not supporting this call in the analyze api and due to the fact that we were returning `null` for the other 2 services, I completely removed the rpc endpoint and dispatched a default action in the component middleware.

### Notes
As this has no backend dependencies I believe it could be merged immediately.

#### Staging Deployment
https://task-deprecate-analytics-start-date-twitter-analyze.dev.buffer.com/


packages/date-picker/reducer.js
```diff
@@ -158,17 +159,7 @@ export default (state = initialState, action) => {
...
-        ...state,
-        loading: true,
-      };
-    case `analytics_start_date_${asyncDataFetchActionTypes.FETCH_FAIL}`:
-      return {
-        ...state,
-        loading: false,
-      };
-    case `analytics_start_date_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
+    case actionTypes.DEFAULT_DATE_PICKER:
```
federicoweber: We are not providing any `result` value on the action so this is effectively just setting `minDate = null` which is already the default we are setting in `initialState` 😉.  Same logic for the presets.

dinostheo: Good one. 
I was only trying to simulate the `null` response of the rpc endpoint without making the request. But I see what you mean now. It's way cleaner that way. I'll change it.

dinostheo: After revisiting the code I noticed that I do need the setDefault flow for reseting the `loading: true` that is set in the initialState. For the rest I can indeed use the initialState.

dinostheo:
I removed the default flow and set the initiaState `loading: false`

PR #516, opened 2019-10-09
[AT-316] Change conditions of disabled options in profile selector

gmzjuliana:
### Purpose
Currently if you filter the list of profiles, if there is only 1, it gets disabled. This changes it so it only gets disabled if it's the currently selected profile


packages/profile-selector/components/ProfileSelectorDropdown/index.jsx
```diff
@@ -163,85 +172,113 @@ const ProfileSelectorDropdown = ({
...
-    );
-    const deactivatedAccount = (disableNetwork === selectedProfile.service);
-    const shouldDisableItems = filteredProfiles.length === 1;
+    const filteredProfiles = filterProfilesByService(
+      profiles,
+      selectedProfile.service
+    ).filter(p => p.username.toLowerCase().match(profilesFilterString));
+    const deactivatedAccount = disableNetwork === selectedProfile.service;
+    const shouldDisableItems =
+      filteredProfiles.length === 1 && //if it's the only one
```
federicoweber: Hey @gmzjuliana do you think we can remove this part of the conditional? I think we always want to disable the selected profile, as clicking on it won't work anyway, regardless of the number of profiles in the list 😉.

packages/profile-selector/components/ProfileSelectorDropdown/index.jsx
```diff
@@ -50,103 +49,112 @@ const DropdownContainer = styled(Dropdown)`
...
   };
 
-  return (<DropdownItem
-    disabled={profile.service === disableNetwork || disabled}
-    key={profile.id}
-    profile={profile}
-    handleClick={onClick}
-    selected={profile.id === selectedProfileId}
-  />);
+  const disabled = profile.id === selectedProfileId;
```
gmzjuliana: heya @federicoweber, I moved this logic into here so we can disable the selected item without depending on it being the only item in the filteredItems array. What do you think?

federicoweber: Great call @gmzjuliana 🙌

PR #517, opened 2019-10-09
kickoff data mock discussion

gmzjuliana:
Heya Fede, I thought we might try reviewing this in github so we can comment back and forth with the code handy.. Here's what I am thinking the data structure will be like, but I'm curious if I'm missing some things 🙈


datamock.js
```diff
@@ -0,0 +1,12 @@
+const initialState = {
+  loading: bool,
+  hasError: bool,
+  primaryMetrics: { totals: [], daily: [] },
+  secondaryMetrics: { totals: [], daily: [] },
```
gmzjuliana: I was thinking that we'd use the storiesDaily endpoint to populate the daily: [] values, what do you think? @federicoweber

federicoweber: That works great @gmzjuliana 🙌, do you want to try and define the daily and totals data structure? I have an hunch we can base it on `compare` 😉 .

datamock.js
```diff
@@ -0,0 +1,12 @@
+const initialState = {
+  loading: bool,
+  hasError: bool,
+  primaryMetrics: { totals: [], daily: [] },
```
federicoweber: Maybe we can just have a `data: {totals: [], daily: []}` field to store all the data, and use the selected metrics to filter those down on `mapStateToProps`?

datamock.js
```diff
@@ -0,0 +1,12 @@
...
+  secondaryMetrics: { totals: [], daily: [] },
+  primaryMetricsLabel: string,
+  secondaryMetricsLabel: string,
+  selectedMetricLabel: string,
```
federicoweber: I think this is the same ar `primaryMetricsLabel` right?

datamock.js
```diff
@@ -0,0 +1,12 @@
+const initialState = {
+  loading: bool,
+  hasError: bool,
+  primaryMetrics: { totals: [], daily: [] },
+  secondaryMetrics: { totals: [], daily: [] },
+  primaryMetricsLabel: string,
+  secondaryMetricsLabel: string,
```
federicoweber: I'm tinkering with the idea of storing the keys in an object
```
selectedMetrics: {
   primary: PRIMARY_METRIC_KEY,
   secondary: SECONDARY_METRIC_KEY,
}
```

and use those keys to filter down daily and totals directly on `mapStateToProps`.

Do you think that could work?

datamock.js
```diff
@@ -0,0 +1,12 @@
...
+  selectedMetricLabel: string,
+  visualizePreviousPeriod: bool,
```
federicoweber: This is acting a bit different in here, as effectively is an option in the secondary metric dropdown, I think we can drop this and just use `selecetedMetrics.secondary`.

🤔 maybe we can extend `selectedMetrics` to have a flag for previous period:
```
selectedMetrics: {
   primary: { key: PRIMARY_METRIC_KEY, isPreviousPeriod: false },
   secondary: { key: SECONDARY_METRIC_KEY, isPreviousPeriod: true},
}
```

datamock.js
gmzjuliana: I updated this partially, but I'd love to understand a bit better how we can filter it on `mapStatetoProps`

datamock.js
gmzjuliana: I added this in but I'm wondering if we do need it since we won't be comparing with the prev period? Maybe I'm missing something there 🤔

datamock.js
gmzjuliana: yup! good catch!

datamock.js
gmzjuliana: that sounds great! 👍

datamock.js
federicoweber: looking at the design specification I think that the second dropdown allows the selection of the previous period, for any of the selected metrics 😉

image
image

federicoweber: For the primary metric you are right we don't need the flag

datamock.js
federicoweber: I'm thinking we can add to small method in the package `index.js` (that is also where the mapping is happening), to filter out the daily and totals based on the selected metrics, so that we could assign those as props those like this:

```
{
    primaryMetric: {
        daily: getMetricDaily(state.storiesDaily.data, state.storiesDaily.selectedMetrics.primary),
        total: getMetricTotal(state.storiesDaily.data, state.storiesDaily.selectedMetrics.primary),
    },
    secondaryMetric: {
        daily: getMetricDaily(state.storiesDaily.data, state.storiesDaily.selectedMetrics.secondary),
        total: getMetricTotal(state.storiesDaily.data, state.storiesDaily.selectedMetrics.secondary),
    }
}
```

How does that sound?

gmzjuliana: ooooh that makes sense, thanks Fede

datamock.js
gmzjuliana: oooh good catch

gmzjuliana: ooo good catch!

gmzjuliana:
heya @federicoweber updating this one with the mock for the data.. sorry for taking a bit longer, I wanted to make sure I really understood 🙈if it feels ok with you, I'll continue on creating the new package tomorrow (wednesday) and we can chat if I'm missing anything on our sync :)

PR #519, opened 2019-10-16
[AT355] Create IG stories breakdown package

gmzjuliana:
### Purpose
create the package where the stories metrics breakdown will live


gmzjuliana:
Heya @federicoweber,

This was meant to be just creating the new package but with the clean up it looks like this:
image

This looks like AT356 too 🙈which we had agreed would be pulling in the resuable components and setting it all up in the right layout with the loading, basic state and no data state

PR #523, opened 2019-10-24
[AT356] Pull in Reusable Components

gmzjuliana:
### Purpose
Creates the backbones for us to work on the instagram metrics breakdown chart. Includes:

- title
- add to report button (no functionality yet)
- layout properly
- error state (this comes built in with the common chart)
- loading state
- basic selects

### Notes
This is what it's currently looking like in storybook: 
image


packages/stories-metrics/components/StoriesMetrics/index.jsx
```diff
@@ -45,23 +68,46 @@ const StoriesMetrics = ({ daily, loading }) => {
     content = <ChartWithFooter daily={daily} />;
   }
 
+  let dropdowns = (
```
federicoweber: I think we don't want to render this for loading and no-data states, as we won't have the selected metrics 😉.

federicoweber:
as for the empty test file, I removed it from a master in https://github.com/bufferapp/buffer-analyze/commit/d4424d25e86936cf4c25bbb7a37f892713b0d3b4, as that was breaking all the tests :).

packages/stories-metrics/components/StoriesMetrics/index.jsx
gmzjuliana: good catch! thanks for the heads up :)

PR #525, opened 2019-10-25
[AT357] Adds custom button select

gmzjuliana:
### Purpose
For IG stories chart, we'll need a new dropdown with a custom button. This sets us up to implement it in our future work

### Notes
Here's what it's looking like in storybook: https://share.getcloudapp.com/NQueYgD4


gmzjuliana:
Heya @federicoweber I'm curious to hear your thoughts as this will end up needing a few more dynamic changes:
- colour of the decorator seems to change in the designs (one is blue and one is pink)
- title will be dependent of current selected metric
- metric in bold will change depending on number

As you'll notice, I've hardcoded these for this iteration but I'm curious if you'd advise adding the dynamic rendering here or later when we're integrating all of this together :) Thanks in advance for your thoughts :)

packages/stories-metrics/components/InfoLabelSelect/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+  font-weight: ${fontWeightBold};
+  font-size: 24px;
+  line-height: 28px;
+  color: #333b43;
+`;
+
+const InfoLabelSelect = ({metrics}) => (
+  <Container>
+    <Select
+      customButton={onButtonClick => (
```
federicoweber: I think this should be defined as a prop 🤔, right?

gmzjuliana: this feels a lot weird 🙈haha, is this sort of what you meant @federicoweber

federicoweber: Haha sorry for the confusion :).

Sorry I had to check the implementation of the `Select` to understand it. So `customButton` is expecting a function and is passing down the event listener as the argument. It makes sense now, it's a bit of an unorthodox implementation ( I would have expected to pass down a `node` with props ), so it threw me off when I was reviewing the code 😅. No need to define it in the props.

gmzjuliana: No worries! I reverted this back :)

packages/stories-metrics/components/InfoLabelSelect/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+        <ButtonContainer onClick={onButtonClick}>
+          <Decorator />
+          <LabelContainer>
+            <Title>Impressions</Title>
+            <Metric>28,426</Metric>
+          </LabelContainer>
+          <ChevronDown />
+        </ButtonContainer>
+      )}
+      items={metrics}
```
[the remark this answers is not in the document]
gmzjuliana: sounds good!

packages/stories-metrics/components/InfoLabelSelect/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+const InfoLabelSelect = ({ metrics, selectedMetric }) => (
+  <Container>
+    <Select
+      customButton={onButtonClick => (
+        <ButtonContainer onClick={onButtonClick}>
+          <Decorator color={selectedMetric.color} />
+          <LabelContainer>
+            <Title>{selectedMetric.title}</Title>
+            <Metric>{numberWithCommas(selectedMetric.number)}</Metric>
```
federicoweber: Hey @gmzjuliana, for this one I think we could use the [TruncatedNumber](https://github.com/bufferapp/buffer-analyze/blob/master/packages/shared-components/TruncatedNumber/index.jsx) component, to ensures that we are using the same formatting logic all around 😉.

gmzjuliana: ah, excellent! Thanks @federicoweber 🌮

PR #526, opened 2019-10-25
[AT358] Remove footer from common chart when needed

gmzjuliana:
### Purpose
Add the option to remove the footer from CommonChart so we can use it for IG Stories Metric Breakdown
### Notes
image


packages/shared-components/CommonChart/index.jsx
```diff
@@ -280,6 +280,7 @@ class Chart extends PureComponent {
       selectedMetrics,
       selectedPreset,
       timezone,
+      withoutFooter
```
gmzjuliana: I normally prefer "truthy" props but we're using commonChart already in two places and I didn't want to break their current implementation. Curious how this feel for you @federicoweber

federicoweber: Good one, personally I will go with `footerless`: I like to use adjectives for this kind of props when possible, and because the missing footer is the variation I prefer to have that as an optional prop.

Other than the name, if we define this prop in `propTypes` and also provide a default value, we won't need to be concerned with breaking the implementation 😉.

PR #527, opened 2019-10-28
[AT359] Create reducer for instagram stories breakdown

gmzjuliana:
### Purpose
define default state and define action for filtering data (select metric)


packages/stories-metrics/reducer.js
```diff
@@ -0,0 +1,92 @@
...
+
+const initialState = {
+  loading: true,
+  hasError: false,
+  data: {
+    totals: [],
+    daily: []
+  },
+  selectedMetrics: {
+    primary: {},
```
federicoweber: Hey @gmzjuliana, as the populated value will be a string do you think we can use `null` or empty string as default values?

packages/stories-metrics/reducer.js
```diff
@@ -0,0 +1,92 @@
...
+function isMetricValid(totals, metricLabel) {
+  const metrics = totals.map(m => m.label);
+  return metrics.indexOf(metricLabel) !== -1;
+}
+
+function getSelectedMetricLabelsOnFetch(state, result) {
+  const totals = result.totals;
+  if (totals.length > 0) {
+    if (
+      state.selectedMetricLabel !== "" &&
```
federicoweber: I don't see `selectedMetricLabel` in our state 😕. Should it be `!state.selectedMetrics.primary`?

packages/stories-metrics/reducer.js
```diff
@@ -0,0 +1,92 @@
...
+      isMetricValid(totals, state.selectedMetricLabel)
```
federicoweber: what is the reason we need this?
Given that we are extracting the metrics from the totals, and we only support one service, my hunch is that we shouldn't really see invalid metrics.

packages/stories-metrics/reducer.js
```diff
@@ -0,0 +1,92 @@
...
+      return state;
+  }
+};
+
+export const actions = {
+  selectMetric(metricLabel) {
+    return {
+      type: `stories_metrics_${actionTypes.SELECT_METRIC}`,
+      metricLabel,
+      isPrimary
```
federicoweber: I think we aren't passing this down 😉

packages/stories-metrics/reducer.js
gmzjuliana: yes! I used an empty string to keep what I had in `getSelectedMetricLabelsOnFetch`

packages/stories-metrics/reducer.js
gmzjuliana: gotcha, thanks for catching this

gmzjuliana:
Added these changes into #539

PR #539, opened 2019-10-31
[AT360] Integrate IG chart into app

gmzjuliana:
### Purpose
using mock data, add the IG chart into the app


gmzjuliana:
Heya @federicoweber :) an extra question on this one.. the tests for the reducer won't pass because of the mocks, what would you recommend for when we merge this one in?

packages/shared-components/CommonChart/components/ChartTooltip/index.jsx
```diff
@@ -106,15 +106,15 @@ const PresetsTooltip = ({
   metricData,
   presetConfig,
   metrics,
-}) => (
+}, wording = null) => (
```
federicoweber: I think we can remove this right?
It wasn't behaving as expected, and I think we only need to pass `wording` down **as a prop** for the custom tooltip.

packages/stories-metrics/components/InfoLabelSelect/story.jsx
```diff
@@ -11,7 +11,7 @@ const metricsWithClickHandlers = metrics.map((metric, index) => ({
     onItemClick: () => console.info(metric.label, 'Clicked'),
 }));
 
-const selectedMetric = {
+const displayedMetric = {
```
federicoweber: ❤️ good call on renaming this

packages/shared-components/CommonChart/components/ChartTooltip/index.jsx
gmzjuliana: yup, thanks for catching this

packages/server/rpc/index.js
```diff
@@ -69,7 +70,7 @@ module.exports = rpc(
   getAccountDetails,
   hashtags,
   aggregate,
-  lastUpdated,
+  //lastUpdated,
```
federicoweber: Hey @gmzjuliana I think we removed this, any reason it has been commented in here?

packages/shared-components/CommonChart/chartConfig.jsx
```diff
@@ -125,9 +125,10 @@ export default () => ({
   xAxis: getXAxis(),
   yAxis: getYAxis(),
   chart: {
-    spacingLeft: 15,
-    marginRight: 55,
+    spacingLeft: 24,
+    spacingRight: 24,
     spacingTop: 40,
+    height: 450
```
federicoweber: 🙌

packages/shared-components/TruncatedNumber/index.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import numeral from 'numeral';
 
-const TruncatedNumber = ({ children, absoluteValue, shorterOption, showPercentSign }) => {
+const TruncatedNumber = ({ children, absoluteValue, shorterOption, showPercentSign, showTwoDecimals }) => {
```
federicoweber: Hey @gmzjuliana how do you feel about always showing two decimals when showing percentages?

packages/stories-daily/components/StoriesDaily/index.jsx
```diff
@@ -0,0 +1,223 @@
...
+
+  dailyWithColor() {
+    const { daily, selectedMetrics } = this.props;
+    return daily.map(item => ({
+      day: item.day.toString(),
+      metrics: item.metrics.map(metric => ({
+        ...metric,
+        color:
+          metric.label === selectedMetrics.primary.label
+            ? twitter
```
federicoweber: it took me a bit to realize this was a color 😄

packages/stories-daily/middleware.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+const formatDataForCSVExport = (
+  { selectedMetrics, daily },
+  { profiles, selectedProfileId }
+) => {
+  const data = {
+    date: [],
+    [selectedMetrics.primary.label]: [],
+    [selectedMetrics.secondary.label]: []
```
federicoweber: 🤔 might be worth returning all the metrics for a given day. What do you think?

packages/server/rpc/index.js
gmzjuliana: sorry just noticed your comments 🙈 I comment this out so I can work without the hot reloader in my local but I just put it back in.. is this meant to be taken out completely?

packages/stories-daily/middleware.js
gmzjuliana: I kinda like keeping it true to what the chart is showing 🙈

packages/shared-components/TruncatedNumber/index.jsx
gmzjuliana: into it! that's a great way to simplify this :)

PR #558, opened 2019-11-25
[AT412] Shopify Totals

gmzjuliana:
### Purpose
Extract TotalsGrid from AggregatesDaily and create a ShopifyTotals component

### Screenshot
image

### Notes
I added this into Instagram/Stories for now so you can see it

#### Staging Deployment
https://at412-shopify-totals-analyze.dev.buffer.com/


packages/aggregates-daily/components/DailyTotalsTable/index.jsx
```diff
@@ -107,8 +26,35 @@ const DailyTotalsTable = ({ metrics, loading, hasError, profiles }) => {
   } else if (Object.keys(metrics).length === 0) {
     content = <NoData />;
   } else {
+    const prettyPrintKey = (key) => {
+      if (key === 'followers'){
```
federicoweber: super small one, I think a switch will look nicer in here 😉.

packages/aggregates-daily/mocks/totals.js
```diff
@@ -1,59 +1,66 @@
...
-        '2019-06-14': 424,
-        '2019-06-15': 424,
-        '2019-06-16': 423,
-        '2019-06-17': 422,
-        '2019-06-18': 425,
-      },
-      total_diff: 2,
+export default [
+  {
+    color: '#fda3f3',
```
federicoweber: He @gmzjuliana I might be missing something but I think we don't need the color right?

packages/shopify-totals/mocks/totals.js
```diff
@@ -0,0 +1,66 @@
...
+      '2019-06-13': 500,
+      '2019-06-14': 350,
+      '2019-06-15': 600,
+      '2019-06-16': 250,
+      '2019-06-17': 10,
+      '2019-06-18': 500,
+    },
+  },
+  {
+    color: '#fda3f3',
```
federicoweber: same thing in here, do we need the color?

PR #559, opened 2019-11-27
[AT413] Top Sources Table

gmzjuliana:
### Purpose
Create a Shopify Top Sources components that sorts based on a selected metric

### Screenshot
image

### Notes
I added this into Instagram/Stories tab so you can see it. The add to report feature doesn't work yet

#### Staging Deployment
https://at413-top-sources-analyze.dev.buffer.com


packages/shared-components/GridList/index.jsx
```diff
@@ -0,0 +1,105 @@
...
+  li {
+    margin: 0;
+
+    :last-child {
+      padding-right: 0;
+    }
+  }
+`;
+
+const LableAndValue = styled.div`
```
federicoweber: i think this should be `LabelAndValue` 😉

packages/shared-components/GridList/mocks/totals.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+  svg {
+    fill: white;
+  }
+`;
+
+export default [
+  {
+    source: "Google",
+    icon: <PlaceholderIcon color="#1DA463"><GooglePlus size="extraLarge" /></PlaceholderIcon>,
```
federicoweber: I think we can remove the icons from the mock right?
We are adding those in `ShopifySources.getIcon` , and ideally mocks should have just raw data :)

packages/shopify-sources/components/ShopifySources/index.jsx
```diff
@@ -0,0 +1,152 @@
...
+import PropTypes from "prop-types";
+import styled from "styled-components";
+import {
+  ChartStateNoData as NoData,
+  ChartStateLoading as Loading,
+  ChartCard,
+  ChartHeader,
+  GridList
+} from "@bufferapp/analyze-shared-components";
+import {
```
federicoweber: Small one, but cause we are going to have quite a few icons in here I think it will be cleaner to do
`import * as icons from "@bufferapp/ui/Icon";`, and access those by source `icons[source]`. How does that sound to you?

packages/shared-components/GridList/mocks/totals.jsx
gmzjuliana: heya @federicoweber is it alright that this would break in storybook?

packages/server/rpc/shopifySources/index.js
```diff
@@ -0,0 +1,97 @@
...
+const rp = require('request-promise');
+const Promise = require('bluebird');
+
+module.exports = method(
+  'shopify_sources',
+  'fetch shopify sources',
+  () => {return new Promise((resolve) => resolve(totals))}
+);
+
+const totals = [
```
federicoweber: Hey @gmzjuliana can we use `./mocks/totals` in here?

packages/shared-components/GridList/mocks/totals.jsx
federicoweber: Hey Juliana, I think we want to have this working in storybook too.
What is the reason stories will break?

federicoweber: Oh I noticed stories are already broken at the moment with `TypeError: this.props.fetch is not a function`.
I think we just need to pass down `fetch` 😉

packages/shopify-sources/middleware.js
```diff
@@ -0,0 +1,90 @@
...
+    });
+  });
+  return data;
+};
+
+export default store => next => action => {
+  // eslint-disable-line no-unused-vars
+  const { dispatch, getState } = store;
+  const state = getState();
+  switch (action.type) {
```
federicoweber: Hey @gmzjuliana I think we want to move the sorting in here.
I noticed that right now, you are handling sorting on the front-end, but that is only taking in account the sources we have already downloaded, so it might not paint the accurate picture. I think we want to have this behave similarly to the sorting in `posts` where we are sending a request to the RPC endpoint containing the metric and the order we are sorting with. 

Note, this will break sorting till we add in APIs, but that's totally ok , as long as we are sending the expected request out.

packages/shopify-sources/components/ShopifySources/index.jsx
```diff
@@ -0,0 +1,164 @@
...
+      return {
+        source: item.source,
+        metrics: [...item.metrics],
+        icon: this.buildIcon(item.source)
+      };
+    });
+
+    const copyOfMetrics = [...metricsWithIcon];
+
+    const sorted = copyOfMetrics.sort(
```
federicoweber: Left more details on the middleware, but I think we want to move this to the back-end

packages/web/components/Stories/index.jsx
```diff
@@ -4,9 +4,11 @@ import { connect } from 'react-redux';
 import Breakdown from '@bufferapp/stories-breakdown';
 import Summary from '@bufferapp/stories-summary-table';
 import StoriesDaily from '@bufferapp/stories-daily';
+import ShopifySources from '@bufferapp/shopify-sources'
 
 const Stories = () => (
   <Fragment>
+    <ShopifySources />
```
federicoweber: let's make sure we are removing this before merging 😉

packages/shared-components/GridList/mocks/totals.jsx
gmzjuliana: gotcha, thanks Fede!

What happens if we don't include the icon is that the component fails when it renders because the icon is getting built and passed in shopify-sources (or wherever else we'll use it in the future). Curious to hear your thoughts on this, not sure how to make it more elegant 🙈

packages/server/rpc/shopifySources/index.js
gmzjuliana: heya @federicoweber would you be up for helping me with the syntax on this one? I tried:
`const totals = require('../../../shopify-sources/mocks/totals')` but I'm getting a build error when I `./dev up analyze` because the totals mock starts with export default? 
image

federicoweber: hey @gmzjuliana, yep that's expected because node uses CommonJS modules.

Another thing, we shouldn't directly import modules from outside of a package, as that will mess up lerna. I think the best option would be to have a copy of the mock in the rpc folder and one in the component.

packages/shared-components/GridList/mocks/totals.jsx
gmzjuliana: just tagging you in the above @federicoweber 🙈

packages/shopify-sources/middleware.js
```diff
@@ -0,0 +1,98 @@
...
+          name: "shopify_sources",
+          args: {
+            profileId: state.profiles.selectedProfileId,
+            startDate: state.date.startDate,
+            endDate: state.date.endDate
+          }
+        })
+      );
+      break;
+    case actionTypes.SELECT_METRIC:
```
gmzjuliana: heya @federicoweber  curious how this one feels now?

packages/server/rpc/shopifySources/index.js
gmzjuliana: aaah gotcha, i'll get this one tomorrow @federicoweber

packages/shared-components/GridList/mocks/totals.jsx
federicoweber: oh ok I see that, and we want to keep the generation of the icons outside of `GridList` cause we are going to use the same grid with the flags 🤔.

That's a good one, ok let's keep it like this, thanks for clarifying.

packages/shopify-sources/middleware.js
```diff
@@ -0,0 +1,98 @@
...
+  const { dispatch, getState } = store;
+  const state = getState();
+  switch (action.type) {
+    case dateActionTypes.SET_DATE_RANGE:
+      if (getState().profiles.selectedProfileId) {
+        dispatch(
+          actions.fetch({
+            name: "shopify_sources",
+            args: {
+              profileId: getState().profiles.selectedProfileId,
```
federicoweber: I think we can use `state` in here 😉, you are already extracting that in line 35.

packages/shopify-sources/middleware.js
federicoweber: looks great 🙌

PR #564, opened 2019-12-11
fix add to report for aggregates dailiy

federicoweber:
### Purpose
To fix rendering of Totals chart in reports

#### Staging Deployment


packages/aggregates-daily/components/DailyTotalsTable/index.jsx
```diff
@@ -38,46 +38,46 @@ function formatDailyData(dailyData, { label, key }) {
   });
 }
 
-export const Table = ({ metrics }) => {
+export const Table = ({ followers, impressions, engagement }) => {
```
gmzjuliana: I'm gonna change this a bit in my PR for Shopify Totals 🙈anything I should keep in mind @federicoweber ?

PR #565, opened 2019-12-12
[AT411] Add Shopify to Sidebar

gmzjuliana:
### Purpose
Adds Shopify to sidebar if not connected so we can show a cross sell page (that I've called ChannelExpansion)

#### Staging Deployment


packages/nav-sidebar/reducer.js
```diff
@@ -1,14 +1,18 @@
 import { actionTypes as dataFetchActionTypes } from '@bufferapp/async-data-fetch';
 import { actionTypes as profileActionTypes } from '@bufferapp/analyze-profile-selector';
 
-const getFirstProfileForService = (service, profiles) => (
-  profiles.find(profile => profile.service === service)
-);
+const getFirstProfileForService = (service, profiles) => {
```
gmzjuliana: my idea here was to create a fake object that we can use to know if there's no shopify page connected. What do you think @federicoweber ?

packages/web/components/ShopifyOverview/index.jsx
```diff
@@ -0,0 +1,26 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import ChannelExpansion from '@bufferapp/channel-expansion'
+
+const ShopifyOverview = ({match}) => {
+  if (match.url === '/shopify/shopify-overview/shopify'){
```
gmzjuliana: here's where we'd catch if Shopify isn't connected and if it is, we would include the revelant components in the else

packages/web/components/ShopifyOverview/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+    return <ChannelExpansion service={"shopify"}/>
```
gmzjuliana: curious if there's another way to pass the service without passing it from props here 🤔

PR #592, opened 2020-01-21
test with separatedProfiles

gmzjuliana:
### Purpose
Separate social profiles so we can aggregate only social profiles

#### Staging Deployment

https://fix-separate-social-non-social-profiles-analyze.dev.buffer.com/


packages/profile-selector/reducer.js
```diff
@@ -13,7 +13,10 @@ export const actionTypes = keyWrapper('PROFILE_SELECTOR', {
 });
 
 const initialState = {
-  profiles: [],
+  profiles: {
```
gmzjuliana: I'm not sure that I've ever seen this kind of declaration in the initialState, is this okay?

gmzjuliana:
eep @federicoweber I gotta fix a bunch of tests but I also gotta run out for today 🙈 If you've got time, I'd love to hear your thoughts on the solution and I'll fix tests tomorrow AM :)

packages/profile-selector/reducer.js
federicoweber: yep that's ok 😉

packages/profile-selector/reducer.js
```diff
@@ -72,12 +75,25 @@ function ensureTimezone(profile) {
   };
 }
 
+function separateProfiles(profiles){
+  const timezoneEnsured = profiles.map(profile => ensureTimezone(profile))
```
federicoweber: We might want to separate this into a different method 😉

packages/profile-selector/reducer.js
```diff
@@ -72,12 +75,25 @@ function ensureTimezone(profile) {
...
+
+  let sortedProfiles = {
+    social: [],
+    nonSocial: []
+  }
+  
+  timezoneEnsured.map(profile => profile.service === 'shopify' ? sortedProfiles.nonSocial.push(profile) : sortedProfiles.social.push(profile))
```
federicoweber: This will work, but I would personally prefer a whitelist approach instead of a blacklist in here so that all non-social profiles are getting filtered by default, something like `const SOCIAL_SERVICES = [`instagram', 'facebook', 'twitter'];`.

What do you think?

packages/server/rpc/profiles/index.js
```diff
@@ -26,6 +28,22 @@ function concatenateShopify(profiles) {
   }]);
 }
 
+function separateProfiles(profiles) {
```
gmzjuliana: the purpose of this is because the profile-selector middleware sometimes receives the profiles as an array of all the   profiles and sometimes well receive them sorted. We chose to move the sorting logic here so it will always receive the same format of information.. we just need to finish implementing it 😅

packages/account/middleware.js
```diff
@@ -15,12 +15,12 @@ export default store => next => (action) => { // eslint-disable-line no-unused-v
       store.dispatch(actions.fetch({
         name: 'get_account_details',
         args: {
-          organizationId: action.result[0].organizationId,
+          organizationId: action.result.social.concat(action.result.nonSocial)[0].organizationId,
```
federicoweber: We chatted about this already, let's abstract this concatenation into a method in `profile-selector` we can use all around

packages/server/rpc/profiles/index.js
federicoweber: This is looking good now 🙌

packages/server/rpc/profiles/index.js
```diff
@@ -35,7 +52,7 @@ module.exports = method(
         const response = await authenticationService.getChannels({
           accountId: req.session.global.accountId
         });
-        return concatenateShopify(response.map(profile => ({
+        return separateProfiles(concatenateShopify(response.map(profile => ({
```
federicoweber: We removed `concatenateShopify` in #589. Before merging this in we will need to rebase it against the last version of `task/shopify-products` 😉

packages/profile-selector/middleware.js
```diff
@@ -102,27 +102,32 @@ function profilesHasData(profiles) {
   return !!profiles.filter(profile => !!profile.last_updated).length;
 }
 
+function concatenateResult(result){
```
federicoweber: Hey @gmzjuliana, do you think we can export this and use it in the other packages instead of redefining it?

Maybe also making the name a bit more explicit `getAllProfilesFromResult` 🤔.

gmzjuliana:
hey @federicoweber there's a test that fails here in `/packages/stories-summary-table` but it's originating from `task/shopify-products`, how does it feel to merge this in and fix that test in that branch?

federicoweber:
> hey @federicoweber there's a test that fails here in `/packages/stories-summary-table` but it's originating from `task/shopify-products`, how does it feel to merge this in and fix that test in that branch?

that is my fault 😅

PR #596, opened 2020-01-29
Task/best time to post parent

gmzjuliana:
### Purpose
a parent branch that we can work off while building best time to post :)

#### Staging Deployment


packages/best-time/middleware.test.js
```diff
@@ -4,21 +4,13 @@ import { actions } from "./reducer";
 
 describe("middleware", () => {
   const next = jest.fn();
-  const profileId = "a-profile-id";
```
gmzjuliana: aaah good catch, thanks for fixing these @federicoweber

federicoweber: sure just some clean up 😉

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,351 @@
...
+          bottom: 0;
+          background-color: white;
+          transition: background-color 0.2s ease;
+        }
+      }
+    `}
+`;
+
+const calculateWhichChildrentoHide = timezone => {
+  // these numbers come from trial and error
```
federicoweber: 😂

packages/best-time/reducer.js
```diff
@@ -0,0 +1,237 @@
...
+    hydratedDays.push({
+      weekday: DAYS_OF_WEEK[index],
+      prediction: getDailyPrediction(createArray(item)),
+      data: createArray(item)
+    });
+  });
+  return hydratedDays;
+}
+
+function mapDataToTimezone(response, timezone) {
```
federicoweber: so nice now, great work 🙌 🌮

packages/best-time/reducer.js
```diff
@@ -0,0 +1,237 @@
...
+        ...initialState,
+        loading: false,
+        hasError: true
+      };
+    case `best_time_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
+      return Object.assign({}, state, {
+        loading: false,
+        ...cleanResponse(action.result, state.timezone)
+      });
+    case profileActionTypes.SELECT_PROFILE:
```
federicoweber: I like this approach to get the timezone, great call 🙌

packages/best-time/reducer.test.js
```diff
@@ -0,0 +1,202 @@
...
+  applyTimezoneOffsetToHours,
+  divideHoursIntoDays,
+  hydrateDaysWithDetails,
+  separateIntoGroups,
+  getHighestDayOfGroup,
+  groupIntoConsummableObject
+} from "./reducer";
+import response from "./mocks/response";
+
+describe("reducer", () => {
```
federicoweber: So many tests 😍

packages/best-time/middleware.js
```diff
@@ -3,7 +3,31 @@ import {
...
+    day: [],
+    time: [],
+    predictedReach: []
+  };
+
+  data.forEach(m => {
+    m.data.forEach(i => {
+      exportData.day.push(m.weekday)
+      exportData.time.push(moment(i.hour, "h").format("ha"));
+      exportData.predictedReach.push(i.prediction)
```
federicoweber: small one @gmzjuliana do you think we can return the percentage instead of the predicted reach?
Thinking about this one, my hunch is that we don't want to show the exact number

gmzjuliana: good call @federicoweber I change it to return the percentage now :)

image

PR #597, opened 2020-01-29
Task/bar chart

gmzjuliana:
#### Staging Deployment


gmzjuliana:
heya @federicoweber wanted to get your thoughts on this one as I'm feeling like I could move to the best_chart component and keep working on the charts there. The reason being that the styling/layout is pretty much done and I'd like to move on to the interaction between the charts. How does that feel?

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,246 @@
...
+      const tooltipLabel = this.createTooltipLabel(percentage);
+      return (
+        <Tooltip customHTML={tooltipLabel} position="top">
+          {singleBar}
+        </Tooltip>
+      );
+    }
+  }
+
+  isLabelShowing(label) {
```
federicoweber: Hey @gmzjuliana, this will work but as we already chatted about using [nth-child](https://css-tricks.com/almanac/selectors/n/nth-child/) to the day of the week and the timezone info. I think we can also use the same approach on `Item` to only show some of the labels.

How does that sound to you?

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,246 @@
...
+  :hover {
+    opacity: 1;
+    cursor: pointer;
+  }
+`;
+
+const Label = styled.label`
+  ${baseFontStyles}
+  
+  color: ${props =>
```
federicoweber: let's add a quick transition between colors :)

packages/best-time/components/BestTime/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+        <React.Fragment>
+          <BarChart
+            metrics={days}
+            type="days"
+            selectedDay={selectedDay}
+            selectNewDay={selectNewDay}
+          />
+          <BarChart
+            metrics={hours}
+            type="hours"
```
federicoweber: small one, but do you think we can use the existence of `selectNewDay` prop to know what kind of chart we are rendering?

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  render() {
+    const { metrics, type, timezone, selectedDay } = this.props;
+    // grab the highestValue so we can calculate the height of the bars as percentages
+    const highestValue = Math.max(...metrics.map(m => m.prediction));
+
+    return (
+      <List>
+        {type === "hours" && <DateTimezoneInfo>{selectedDay}</DateTimezoneInfo>}
```
federicoweber: I think we can just check for the presence of the `selectedDay` prop right?
`{selectedDay && <DateTimezoneInfo>{selectedDay}</DateTimezoneInfo>}`.

I left a similar comment about removing the `type` prop and check for the existence of the `selectNewDay` prop, the reason for that is that this will make the component a bit more flexible, having the individually use any of those options. 

What do you think about it @gmzjuliana?

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,246 @@
...
+              >
+                {this.createBar(metric, height)}
+                <Separator />
+                {label}
+              </Label>
+            </Item>
+          );
+        })}
+
+        {type === "hours" && (
```
federicoweber: Same thing in here `{timezone && <DateTimezoneInfo right>{timezone}</DateTimezoneInfo>}` 😉

packages/best-time/reducer.js
```diff
@@ -0,0 +1,69 @@
...
+  baseReach: getBaseReach(response)
+};
+
+function getBaseReach(response) {
+  return Math.min(
+    ...response.map(x => Math.min(...x.data.map(m => m.prediction)))
+  );
+}
+
+function findNewHours(state, newDay) {
```
federicoweber: to keep names consistent between methods let's use `getHours` 😉

packages/best-time/reducer.js
```diff
@@ -0,0 +1,69 @@
...
+        ...initialState,
+        loading: false,
+        hasError: true
+      };
+    case `best_time_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
+      return {
+        loading: false,
+        selectedDay: "Monday",
+        data: response,
+        hours: response[0].data,
```
federicoweber: this will work, but to keep it consistent all around I would prefer to use `getHours()`, it will make things easier to read 😉.

We can modify that method signature to accept `data` instead of the `state`, so that it could work in both cases.
```
getHours(response, 'Monday');
getHours(state.data, action.newDay);
```

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,272 @@
...
+          width: 100px;
+          height: 16px;
+          position: absolute;
+          left: 0;
+          bottom: 0;
+          background: white;
+        }
+      }
+
+      &:nth-child(n + 20):hover {
```
federicoweber: Do you think we can calculate the width and the number of children that needs to hide the label based on the length of the timezone string 🤔?

I was checking the possible values in https://en.wikipedia.org/wiki/List_of_tz_database_time_zones, and it seems there is a huge variations in lengths, from 3 chars `UTC`, up to 32 `America/Argentina/ComodRivadavia`.

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,272 @@
...
+    width: 100%;
+    border-bottom: 2px solid #2b6af6;
+    padding-bottom: 8px;
+    position: absolute;
+    bottom: 24px;
+    z-index: 1;
+  }
+`;
+
+const Item = styled.li`
```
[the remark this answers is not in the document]
gmzjuliana: heya @federicoweber I'd love to check this one with Tom while James is away.. I just remembered that I had mentioned this doing our Analyze weekly and I had it previously working with a larger hover area and one of the comments was to have it only enabled on click.. we could make the min-height taller also.. what do you think?

packages/best-time/reducer.test.js
```diff
@@ -0,0 +1,44 @@
...
+      hasError: false,
+      metrics: []
+    };
+  });
+
+  it("should initialize default state", () => {
+    expect(reducer(undefined, { type: "INIT" })).toEqual(initialState);
+  });
+
+  it("it should add labels on FETCH_SUCCESS", () => {
```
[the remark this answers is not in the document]
gmzjuliana: nope, I'll take it out for now :)

packages/best-time/components/BarChart/index.jsx
```diff
@@ -0,0 +1,300 @@
...
+  }
+  if (timezone.length >= 7 && timezone.length <= 12) {
+    return 23;
+  }
+  if (timezone.length >= 13 && timezone.length <= 18) {
+    return 22;
+  }
+  if (timezone.length >= 19 && timezone.length <= 24) {
+    return 21;
+  } else return 20;
```
gmzjuliana: curious to get your thoughts on this as a solution to hide just the right amount depending on the timzone.lenght @federicoweber

packages/best-time/components/BarChart/index.jsx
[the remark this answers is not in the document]
federicoweber: Good one @gmzjuliana, personally I will prefer to have the active area span the entire height, for both hover and click, I think it makes it a bit easier to navigate, but admittedly I'm more at ease with a keyboard than a mouse 😄. 

Will defer to @tomredman

packages/best-time/components/BarChart/index.jsx
federicoweber: Hey @gmzjuliana I think this will work great 🙌. 
The only change I will make is to use a `switch` instead of those long chain of `if` 😉.

federicoweber:
Hey @gmzjuliana, a small one 🙂. 
I noticed that after the click we are getting a squared outline around the button, my hunch is that it's caused by the `focus` state. Can you please take a look a that?

Thanks 🙏!

hover

packages/best-time/components/BarChart/index.jsx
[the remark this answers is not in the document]
tomredman: @gmzjuliana @federicoweber I'd be happy to keep the hover area the full height of the area, just to make that a little easier to navigate. I'm good with the min-height but don't feel strongly either way – feel free to use your best judgment on that!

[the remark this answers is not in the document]
gmzjuliana: sounds good, I'll play around with this today 😁 thanks pals!

PR #602, opened 2020-02-10
fix daily aggregate imports in reports

gmzjuliana:
### Purpose
Got this report from Adam today and on investigating, it's because of faulty imports: https://buffer.slack.com/archives/C0DLE125Q/p1581347717392700

#### Staging Deployment


gmzjuliana:
I didn't merge this because it does not render in reports... more info here: https://buffer.atlassian.net/browse/AT-484

PR #605, opened 2020-02-26
Creates a stacked chart for Campaign Posts Metrics

gmzjuliana:
### Purpose
Part of Campaigns is including these two new graphs:
image
image

#### Staging Deployment


packages/campaign-posts/components/StackedChart/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+function prepareSeries(dailyMetrics) {
+  const series = [];
+  let color = "#9B9FA3";
+  let metricIndex = 0;
+  let lineColor = "#9B9FA3";
+
+  function mapDataToPoint(dailyEntry) {
+    let value = 0;
+    const metric = dailyEntry.metrics[metricIndex];
```
federicoweber: This feels a bit brittle as it's relying on `metricIndex` getting properly set after the push to series.

A slightly better approach could be to use `series.length` directly, as that is what we are really looking for, but ideally I would prefer to avoid this kind of coupling between methods, as it could be hard to follow and easy to break 😉.

packages/campaign-posts/components/StackedChart/index.jsx
```diff
@@ -0,0 +1,141 @@
...
+  data: null,
+};
+
+const colorMap = {
+  facebook: '#90D1F9',
+  twitter: '#7391CD',
+  instagram: '#D15FA0',
+}
+
+function prepareSeries(dailyMetrics) {
```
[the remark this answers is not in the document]
federicoweber: This is a good point @gmzjuliana, we are really doing lot's of work in here, and the various methods are quite interwoven, relying on semi-global variables.
   
Maybe we can abstract this in smaller, testable functions 🤔
- one method to get the basic points `mapDataToPoint(dailyMetrics, metricLabel)`
(where `metricLabel` replaces `metricIndex`, and it's used to get the metrics whose label match.
- one method fill in a series for one metric with the appropriate config and points `makeMetricSeries(dailyData, metricLabel)`
- one method to get all the serieses `mapDailyIntoSerieses(dailyData, metricLabels)`

packages/shared-components/utils/chartUtils/chartConfig.jsx
```diff
@@ -0,0 +1,298 @@
...
+    style: {
+      'font-size': '0.875rem',
+      'font-weight': '400',
+      'font-family': 'Roboto, sans serif',
+      'white-space': 'nowrap',
+    },
+  },
+});
+
+export const getYAxis = (stacked) => {
```
federicoweber: let's break this down into two `getYAxis`, and `getYAxisStacked`

packages/shared-components/utils/chartUtils/chartConfig.jsx
```diff
@@ -0,0 +1,256 @@
...
+  }
+  min -= (min / 100) * bottomPaddingPercentage;
+  max += (max / 100) * topPaddingPercentage;
+  return {
+    min,
+    max,
+  };
+}
+
+export const setYAxisScale = ([primarySeries, secondarySeries, ...otherSeries], xAxis, yAxis) => {
```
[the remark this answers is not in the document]
federicoweber: @gmzjuliana do you think we can simplify this a bit?
maybe one  small method to get `scaleDifference`, and one to decide if we need one or two scales based on that 🤔

packages/shared-components/utils/chartUtils/chartConfig.jsx
```diff
@@ -0,0 +1,298 @@
...
+}
+
+export const fadeColor = (color, opacity) => {
+  const highchartColor = ReactHighcharts.Highcharts.Color(color);
+  return highchartColor ?
+    highchartColor.setOpacity(opacity).get('rgba') :
+    color;
+}
+
+export default (tooltipgetter, stacked = false, type = 'areaspline') => {
```
federicoweber: Super small one, let's change `tooltipgetter` into `tooltip`. Highchart is already expecting a method there.
Also given that this is the only one without a default and we might find the need to hide the tolltip for some series, it might be worth setting the default to  something like this `(() => (null))`.

packages/shared-components/utils/chartUtils/chartConfig.jsx
gmzjuliana: What do you think about this way? Both are copying the basic config object so I thought it might be cleaner to have just one copy of it and copy it as many times as we need. More in the following comment 👇

packages/shared-components/utils/chartUtils/chartConfig.jsx
```diff
@@ -0,0 +1,255 @@
...
+        'font-size': '0.875rem',
+        'font-weight': '400',
+        'font-family': 'Roboto, sans serif',
+      },
+    },
+  }
+  if (stacked){
+    return [{...basicConfig},{...basicConfig},{...basicConfig}]
+  } else
+  return [{...basicConfig},{...basicConfig}]
```
gmzjuliana: This doesn't feel very flexible but I also couldn't think of a chart with 4 or more data sets.. curious how you feel about this one though!

packages/shared-components/utils/chartUtils/chartConfig.jsx
```diff
@@ -0,0 +1,255 @@
...
+  }
+  return scaleDifference;
+};
+
+export const setYAxisScale = ([primarySeries, secondarySeries, ...otherSeries], xAxis, yAxis) => {
+  // we are using two scales only if there is a siginificative difference in scale
+  const yAxisMax = Math.max.apply(null, reduceSeriesToValues([primarySeries]));
+  let usesTwoScales = false;
+
+  const scaleDifference = getScaleDifference(secondarySeries, yAxisMax);
```
gmzjuliana: I split this into one instead of two, curious how this feels now? I was a bit hesitant to split the usesTwoScales into another function because on line 185 or 187 we're modifying the secondarySeries.yAxis.. but if that doesn't need to be coupled here, I'm happy to move it :)

packages/campaign-posts/components/StackedChart/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+  dailyMetrics.map(x => {
+    innerSeries.push({
+      y: x.metrics[index].value,
+      x: getTimeStamp(x.day)
+    }) 
+  })
+  return innerSeries;
+}
+
+function prepareSeries(dailyMetrics) {
```
gmzjuliana: Did a bit of an overhaul with this one 😁 what do you think?

packages/shared-components/utils/chartUtils/chartConfig.jsx
federicoweber: hey @gmzjuliana, given that we are effectively returning the same config multiple times, do you think we can just send the number of series we want instead of `stacked`? Maybe we can default that to 2 so that the normal behavior is not altered.

packages/shared-components/utils/chartUtils/chartConfig.jsx
federicoweber: feels good to me 🙂

PR #611, opened 2020-03-04
Fix the analyze-api address

dinostheo:
### Purpose
The service have changed it's name in our cluster.
It is better to rename the service, but due to the fact that I'm not aware of unexpected side-effects of it. I fix it here temporary to see if the client is recovering. 

#### Staging Deployment


dinostheo:
Changed in the api side

dinostheo:
Still relevant

PR #618, opened 2020-03-06
creates posts timeline

gmzjuliana:
### Purpose
Creates a posts timeline component that will go into the Campaign Posts Metrics component

### Notes
I think we can tweak the spacing of the posts once we start joining the components together.. maybe a calculation based on the number of days in the campaign? Let me know if you think it's a good idea to start adding that in now though!

#### Staging Deployment


packages/campaign-posts/components/PostsTimeline/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+    height: 150px;
+    content: '';
+    display: block;
+    background: transparent;
+    position: absolute;
+    top: -140px;
+  }
+`;
+
+const List = styled.ul`
```
federicoweber: I think this can be a `ol`, we are sorting by dates 😉.

packages/campaign-posts/components/PostsTimeline/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+          if (daily.posts.length) {
+            return (
+              <Item key={index} position={index}>
+                <Tooltip
+                  customLabel={
+                    <TimelineTooltip posts={daily.posts} day={daily.day} />
+                  }
+                  position="top"
+                >
+                  <Post onClick={() => {console.log('hello', index)}}/>
```
federicoweber: let's pass this down as a prop 🙂

packages/campaign-posts/components/TimelineTooltip/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+            <Text normal>
+              {channel.details.length === 1 ? "post" : "posts"}
+            </Text>
+          </Item>
+        ))}
+      </List>
+      <Container>
+        <Button
+          type="secondary"
+          onClick={() => {console.log('hey!')}}
```
federicoweber: let's pass it down as a prop, so we can make sure we are invoking the same event, as when clicking on the entire tooltip.

PR #622, opened 2020-03-11
fix glitch caused by UI/Tooltip

federicoweber:
### Purpose
This fixes a glitch that was happening when using best-time tooltips, where `reach-ui/reach-portal` was momentarily glitching page sizing causing scroll bars to appear.

#### Staging Deployment


packages/server/index.js
```diff
@@ -248,16 +248,27 @@ app.get('/report_to_pdf', (req, res) => {
...
+`;
+
 app.get('*', (req, res) => {
   let parsedHtml;
   if (req.path.includes('/export/')) {
     parsedHtml = html
       .replace('{{{iterateScript}}}', '')
-      .replace('{{{beamerScript}}}', '');
+      .replace('{{{beamerScript}}}', '')
+      .replace('{{{appStyle}}}', '');
```
federicoweber: we are removing this in here otherwise charts will be hidden

PR #626, opened 2020-03-16
Campaigns Content Modal

gmzjuliana:
#### Staging Deployment


gmzjuliana:
hey @moreofmorris would you be up for taking a look at the Card component? I'm figuring out how to put it in the modal 😅 but the layouts are there :)

moreofmorris:
This is look good @gmzjuliana 👍I would say the order of the metrics are important. They need to be easily scannable and pretty consistent between the channels, especially when flicking between posts inside the modal. I was curious, why is this called a card when it's a modal? 🤔Also, it the code seems to sit inside a folder that refers to "campaigns" but I wonder if this should be sitting in something more general like "content-modals" or something like that? These modals will eventually support Stories and will be available everywhere in Analyze and so aren't just for Campaigns (though they are for now) 😊 It would be good too to restrict the height of the body text to a certain number of lines to keep the modal size tight. Showing three lines maximum feels good there but curious to see how we'd do that. I designed it with a slight gradient fade after so many lines. Anything we can do to keep these modals tight and nicely in shape would be great! All looking great though, Juliana, as usual! ❤️I'm working on the posts without images as we speak, so will update you soon 👍

gmzjuliana:
Hey @federicoweber, would you be up for giving me some advice for this that James mentioned:

> Also, it the code seems to sit inside a folder that refers to "campaigns" but I wonder if this should be sitting in something more general like "content-modals" or something like that? These modals will eventually support Stories and will be available everywhere in Analyze and so aren't just for Campaigns (though they are for now) 😊 

A few weeks ago we discussed that we should keep components inside of their modules until we need to abstract them to use them elsewhere. I remember we mentioned that it's better to avoid abstracting too early so that we can move faster and keep things simple. I'd lean towards not abstracting yet because of that, but wondered what your hunch on this is?

federicoweber:
> Hey @federicoweber, would you be up for giving me some advice for this that James mentioned:
> 
> > Also, it the code seems to sit inside a folder that refers to "campaigns" but I wonder if this should be sitting in something more general like "content-modals" or something like that? These modals will eventually support Stories and will be available everywhere in Analyze and so aren't just for Campaigns (though they are for now) 😊
> 
> A few weeks ago we discussed that we should keep components inside of their modules until we need to abstract them to use them elsewhere. I remember we mentioned that it's better to avoid abstracting too early so that we can move faster and keep things simple. I'd lean towards not abstracting yet because of that, but wondered what your hunch on this is?

Hey @gmzjuliana, I think James has a good point here. The component itself is quite independent from campaigns, and we already know that we are going to reuse this for stories soon, so I will go with a more generic name, and structure.

_In general, I think we don't want to abstract for the unknowns but it's always good to consider all the knowns, to avoid rework 😉._

packages/shared-components/ContentModal/components/Card/index.jsx
```diff
@@ -0,0 +1,230 @@
...
+        singleStat === "url_clicks"
+          ? "Clicks"
+          : singleStat.split("_").join(" "),
+      value: stats[singleStat]
+    };
+    cleanedMetrics.push(metricInfo);
+  });
+
+  return cleanedMetrics;
+}
```
gmzjuliana: this mapping will move to the RPC once that's ready to be worked on

gmzjuliana:
hey @moreofmorris curious if you're up for another check of the Card component? I moved that into `shared-components` and you'll see multiple stories in the ContentModal component: https://share.buffer.com/z8uXo8pA

moreofmorris:
It's coming along nicely @gmzjuliana 👍Is there anything we can do to keep things a little more consistent between the types. For example, things wiggle about a bit when going between text with and without emojis. Also that text is feeling a little squashed, I wonder if we could increase the line height there to give it more some more space? The icon next to the Last Updated seems a little off and feels like it should be a little more vertically centered to the text next to it. Also, I wonder if that Last Updated text could be brought down to sit to the bottom of the modal so it's baseline sits at the same level as the text inside the View Post On Network button. Not sure if that makes sense though? 😓Happy to also jump in and make the changes too if that helps? Great work as always, Juliana! 🙌

gmzjuliana:
aw thanks @moreofmorris ! Just added those changes in :)

packages/shared-components/ContentModal/components/Card/index.jsx
```diff
@@ -0,0 +1,256 @@
...
+            />
+          </Link>
+        </LinkPosition>
+      </Divider>
+      <TextContentDivider>
+        <HeadingDivider>
+          <Text color="outerSpace" weight="semi-bold">
+            {moment.unix(post.date).format("D MMMM, YYYY h:mma")}
+          </Text>
+          <ProfileLegend profile={profile} />
```
gmzjuliana: hey @federicoweber I wanted to point this one out because ProfileLegend expects the full profile object like:

```
const facebookProfile = {
  id: "123",
  avatarUrl:
    "https://pbs.twimg.com/profile_images/899569910128553985/89LaWvpV_normal.jpg",
  service: "facebook",
  timezone: "Europe/London",
  username: "@moreofmorris"
};
```

I'm not sure that we should use it and also, it made me realize that we'll need to have the username as well. I think it might make sense to circle back to this when I work on the reducer cause I can see then if we can create a map with the campaign profiles. What do you think?

yarn.lock
```diff
@@ -341,10 +341,10 @@
   resolved "https://registry.yarnpkg.com/@bufferapp/shutdown-helper/-/shutdown-helper-0.2.0.tgz#d43bbba845e30e411dda76a7b7629e11160e5395"
   integrity sha1-1Du7qEXjDkEd2nant2KeERYOU5U=
 
-"@bufferapp/ui@^5.27.0":
-  version "5.27.0"
-  resolved "https://registry.yarnpkg.com/@bufferapp/ui/-/ui-5.27.0.tgz#790e166dc14dc0ade6901a955aafd71d265e0768"
-  integrity sha512-Ybpqt3jU6+6Vq6lAhGyDKjP2fy65CXDBQBvodwGJLkg0VtpHFM6qwiudM5dO44G7ELsENz2mk36QeVU+jRM/fA==
+"@bufferapp/ui@^5.32.0":
```
gmzjuliana: just leaving a note here that v5.31 breaks storiesDaily, I'll chat with Maya about this!

packages/shared-components/ContentModal/components/Card/index.jsx
federicoweber: hey @gmzjuliana, good catch, I noticed this when I was working on the posts breakdown, and Dinos added in a `profile` attribute in [campaign posts](https://paper.dropbox.com/doc/Analyze-Campaigns-technical-specs--AwaoShyvOt1C8D4f2DUF1wP3Ag-4Pgd64eXrqIkFjGulDpXH) 😉

PR #632, opened 2020-03-24
test inerting the focus of the modal

gmzjuliana:
hey @federicoweber I wanted to ask you for some advice for locking the focus inside of the modal so that keyboard users won't get lost in the content behind the modal:

I'm a bit stumped because it feels like there's two ways to dismiss the modal: from the app and using `dismiss` in `Modal.jsx`.. I was thinking of using something like this [which I grabbed from this article](https://www.smashingmagazine.com/2014/09/making-modal-windows-better-for-everyone/)

the ref I put in Modal isn't catching and I'm not sure why. I'm curious 
1. what you think of this approach (maybe this should only happen in BDS and not in each app every time we need to use a modal
2. if you have any other ideas for making this work :)


federicoweber:
Hey @gmzjuliana, this is a great question. I left some comments in the modal PR about using `dismiss`, also for the `X` button.
My hunch is that we want to have the modal handle itself, and use a callback to cleanup any app-specific states. 

How does that sound to you?

PR #635, opened 2020-03-27
get all the components in the same place

gmzjuliana:
### Purpose
image

Getting all the components in the same place for the layout

#### Staging Deployment


gmzjuliana:
Heya @federicoweber I've had 3 issues here that'd I'd love some help with if possible :)

1. how to get the posts lined up with the dates of the chart
2. opening the modal from the tooltip on click (we're using another button there)
3. keeping the modal open even when the user isn't hovering on the tooltip

Just writing these makes me think that maybe that's something we should handle on the parent but I'd be nice to chat about it, if you're up for it :)

federicoweber:
> Heya @federicoweber I've had 3 issues here that'd I'd love some help with if possible :)
> 
> 1. how to get the posts lined up with the dates of the chart
> 2. opening the modal from the tooltip on click (we're using another button there)
> 3. keeping the modal open even when the user isn't hovering on the tooltip
> 
> Just writing these makes me think that maybe that's something we should handle on the parent but I'd be nice to chat about it, if you're up for it :)

We chatted about this, just leaving some notes for transparency 🙂.
1. that was a fun one, super happy how it turned out with 🙌 
2., 3. We are going to move the modal outside of the tooltip, and trigger it from the reducer.

PR #642, opened 2020-04-02
Swap out Add To Report button for one in BDS

moreofmorris:
### Purpose
To swap out the Add To Report button to use the BDS button instead.

### Notes
This does not change the text or buttons inside the modal, yet.

### Staging Deployment
https://task-update-add-to-report-analyze.dev.buffer.com/


moreofmorris:
Yo @msanroman I've added you to this, I hope that's ok. It looks ok from my side but could do with a second pair of eyes 👀Is there anything else I need to consider before I merge? 🤔

PR #650, opened 2020-04-03
Campaigns/post timeline tweaks

gmzjuliana:
### Purpose
- updates Bufferui to pull in Modal changes:
-- makes the background overlay darker
-- moves close button closer to the modal content
-- focuses the modal after open so that ESC works right away
-- removes buttons and nav in the carousel if there’s only 1 item
-- adds click anywhere outside the modal to close for modals with closeButton
-- adds opacity option to tooltip

- changes button in tooltip for note, adds animations to buttons
- keeps same position for modal close button when only 1 post

#### Staging Deployment


moreofmorris:
Hey there @gmzjuliana this is looking lovely and so much better overall, thank you. Just some small ones that I'm picking up here. With the close now so much closer to the modal (which is nice) let's just push that a little further and be more flush with the top right of the modal box as that was actually an earlier design anyhow so feels good now seeing it here, bit like this: https://share.buffer.com/jkuKxbEY and also, I just noticed that when the dot enlarges on hover it seems to enlarge by width and height but it's not enlarging perfectly centered, if that makes sense. Maybe the `x` and `y` coords need to be adjusted too to ensure it's enlarging nicely from the center? Hope that makes sense. These changes are all so much better. Thank you for your hard work! 🙌

gmzjuliana:
heya @moreofmorris, I merged this in with the changes you requested but if you think we need to tweak it a bit more, let me know! 😁

PR #654, opened 2020-04-07
Campaigns UX improvements

gmzjuliana:
### Purpose
fixes the following: 
- If campaign date range is within same month could we shorten the data like they do in Publish? [Ref]
- View post on network button broken in modal [ref]
#### Staging Deployment


packages/campaigns/components/Header/index.jsx
```diff
@@ -123,6 +123,19 @@ const removeEmoji = (name) => {
...
+const isSameMonth = (start, end) => {
+  const startMonth = moment(start).month();
+  const endMonth = moment(end).month();
+  return startMonth === endMonth;
+};
+
+const isSameYear = (start, end) => {
+  const startMonth = moment(start).year();
+  const endMonth = moment(end).year();
+  console.log(startMonth === endMonth);
```
federicoweber, line 135: I think this is a left over 😉

packages/campaigns/components/Header/index.jsx
```diff
@@ -123,6 +123,19 @@ const removeEmoji = (name) => {
   );
 };
 
+const isSameMonth = (start, end) => {
```
federicoweber, line 126: hey @gmzjuliana this works great, and _no need to change_, but it's worth noting that Moment has a convenient [isSame ](https://momentjs.com/docs/#/query/is-same/)helper method 😉

gmzjuliana: ooh thanks for letting me know @federicoweber :)

PR #655, opened 2020-04-08
adds prettyPrinting to RPC calls

gmzjuliana:
### Purpose
Following this thread: https://threads.com/34376598096 we can pretty print out RPC calls so they're easier to debug and easier to track in DataDog

#### Staging Deployment


package.json
```diff
@@ -93,6 +94,7 @@
     "webpack-merge": "^4.2.1"
   },
   "dependencies": {
+    "@bufferapp/async-data-fetch": "^2.0.0",
```
gmzjuliana, line 97: I added async-data-fetch here and removed it from all the other packages, using the same reasoning as the change we recently made for BDS: easier to keep track if it's in one place :)

packages/server/services/authenticationService.js
```diff
@@ -1,5 +1,10 @@
 // todo: move this file to a package in its own repo or in core-authentication-service
```
gmzjuliana, line 1: this is what Hamish advised in the *One Caveat* part of his thread

